PR #135, opened 2018-03-30
Show star in the settings dialog

aminroosta:
@brucebinary Please review


src/store/CategoricalDisplayStore.js
```diff
@@ -221,8 +225,22 @@ export default class CategoricalDisplayStore {
         const layout = this.context.stx.layout;
         layout.favorites[this.favoritesId] = toJS(this.favoritesCategory.data);
```
aminroosta: @brucebinary Just storing `item.itemId` should have been enough.

brucebinary: @aminroosta That works. Guess we don't need to separate favorites out after all. Looks good to merge (LGTM)

PR #149, opened 2018-04-03
For favorites, just storing the itemId should be enough

shayanypn:
@aminroosta  please check the issue 
I am not sure if  the favories use anywhere else in the project, if it is.
I should check it before merging the codes

### DIFF
--- a/src/store/CategoricalDisplayStore.js
+++ b/src/store/CategoricalDisplayStore.js
@@ -233,7 +233,7 @@ export default class CategoricalDisplayStore {
     }
     setFavorite(item) {
         if (this.favoritesMap[item.itemId]) {
-            this.favoritesCategory.data = this.favoritesCategory.data.filter(x => x && x.itemId !== item.itemId && x !== item.itemId);
+            this.favoritesCategory.data = this.favoritesCategory.data.filter(favItem => favItem && favItem.itemId !== item.itemId && favItem !== item.itemId);
             delete this.favoritesMap[item.itemId];
         } else {
             this.favoritesCategory.data.push(item);
--- a/src/store/CategoricalDisplayStore.js
+++ b/src/store/CategoricalDisplayStore.js
@@ -131,23 +131,21 @@ export default class CategoricalDisplayStore {
             items = (filteredItems && filteredItems[0]) ? filteredItems[0].data : [];
 
         if (this.favoritesId) {
-            const oldFavoritesCategory = toJS(this.favoritesCategory);
-            let newFavoritesCategory = [];
+            const favsCategory = toJS(this.favoritesCategory);
+            let favsCategoryItem = [];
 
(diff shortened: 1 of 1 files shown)


src/store/CategoricalDisplayStore.js
```diff
@@ -215,15 +233,15 @@ export default class CategoricalDisplayStore {
     }
     setFavorite(item) {
         if (this.favoritesMap[item.itemId]) {
-            this.favoritesCategory.data = this.favoritesCategory.data.filter(x => x.itemId !== item.itemId);
+            this.favoritesCategory.data = this.favoritesCategory.data.filter(x => x && x.itemId !== item.itemId && x !== item.itemId);
```
aminroosta: nitpick:  please don't use single letter variables like `x` :-)
If you need to shorten the name use 3letter variables like:
`rows.forEach(row => { });` or `favoritesCategory.data.forEach(cat => { })` or `favorites.forEach(fav => { })`.

And if you really need single letter variables which are handy in lambdas use the first letter like `rows.forEach(r => { })` or `favorites.forEach(f => {})`

src/store/CategoricalDisplayStore.js
```diff
@@ -215,15 +233,15 @@ export default class CategoricalDisplayStore {
...
             delete this.favoritesMap[item.itemId];
         } else {
             this.favoritesCategory.data.push(item);
             this.favoritesMap[item.itemId] = true;
         }
 
         const layout = this.context.stx.layout;
-        layout.favorites[this.favoritesId] = toJS(this.favoritesCategory.data);
+        layout.favorites[this.favoritesId] = toJS(this.favoritesCategory.data).filter(item => item).map( item => typeof item === 'string' ? item : item.itemId);
```
aminroosta: @shayanypn why do we need `.filter(item => item)` ?

shayanypn:
@aminroosta we also has favorites in comparsion, but as I change the `caegoryStore` so both of  them(studies and comparsion) works well.

src/store/CategoricalDisplayStore.js
shayanypn: @aminroosta is some cases (ex, user work with old version which store whole object) by adding new fav, and storing it, the code generate some `null` item, that make store crash. just to prevent this crash

src/store/CategoricalDisplayStore.js
shayanypn: sure.
should I modify this part?

src/store/CategoricalDisplayStore.js
aminroosta: @shayanypn Ok, that's good

src/store/CategoricalDisplayStore.js
aminroosta: @shayanypn yes please :-)

PR #152, opened 2018-04-03
Bug fix: The active symbol changer doesn't work on mobile devices (iOS and Android)

brucebinary:
To make touch events work in mobile devices, all events inside `chartContainer` will need `stxtap` event instead of React's `onClick` to work.


src/store/MenuStore.js
```diff
@@ -31,9 +31,7 @@ export default class MenuStore {
     }
 
     @action.bound onTitleClick(e) {
-        /* TODO: why stopPropagation() is not working ಠ_ಠ */
-        // e.stopPropagation();
-        e.nativeEvent.isHandledByDialog = true;
+        e.stopPropagation();
```
brucebinary: @aminroosta now that we use a native event instead of React's event, this should work.

aminroosta:
@brucebinary the `ref={}` is called multiple times during component re-render, I'm not sure if the `stxtap(handler)` will be called twice or not, could you check?

src/store/utils.js
```diff
@@ -50,3 +50,10 @@ export function downloadFileInBrowser(filename, content, type) {
         document.body.removeChild(link);
     }
 }
+
+export function stxtap(el, func) {
+    if (el) {
+        CIQ.installTapEvent(el);
```
aminroosta: What is `installTapEvent` really doing? Can we do it ourself or is it dependent on the internals of the chartiq librarry?

brucebinary: @aminroosta ah shucks. The `ref` is called multiple times. Will have to add a check for that. Will ping you when I'm done

I actually add `installTapEvent` because without it `stxtap` event doesn't get fired. This wasn't the case before (28 Mar) so I wasn't sure what happen. In the docs (http://documentation.chartiq.com/CIQ.html#.installTapEvent) it writes: 

_Converts an object to emit "stxtap" events. This uses CIQ#safeClickTouch. You should use addEventListener("tap") to receive the events._

Of course, "tap" doesn't work. We need to use "stxtap"

src/store/utils.js
```diff
@@ -50,3 +50,10 @@ export function downloadFileInBrowser(filename, content, type) {
         document.body.removeChild(link);
     }
 }
+
+export function stxtap(el, func) {
+    if (el && !el.safeClickTouchEvents) {
```
brucebinary: @aminroosta fixed. `safeClickTouchEvents` is appended to the element via `CIQ.safeClickTouch`, which is called in `CIQ.installTapEvent`

PR #155, opened 2018-04-04
A better UI when Market is closed

binarydevteam:
@brucebinary  display closed icon beside of being disable in markets.


src/components/Icons.jsx
```diff
@@ -4,6 +4,7 @@ import Add from '../../sass/icons/chart settings/zoom-in/ic-zoomin-light.svg';
 import AddThin from '../../sass/icons/add/ic-add.svg';
 import Candle from '../../sass/icons/chart settings/chart types/candle/ic-candle-normal.svg';
 import Close from '../../sass/icons/close/ic-close.svg';
+import ClosedMarket from '../../sass/icons/close/ic-market-closed-line.svg';
 import Comparison from '../../sass/icons/chart settings/comparison/ic-comparison-normal.svg';
```
brucebinary: where is `ic-market-closed-line.svg`? I didn't find it in my source code. Also I think you could just css for this. It's much easier to change later on.

binarydevteam: yeah you are right, I forgot to add the SVG file in the push.
I try to follow the prev codes, but I also can attached it as style icon only on css.
which one you prefer ?

brucebinary:
@binarydevteam  Cool it works... I just think that the SVG file should be replaced with a `div` block with CSS. Saves space and easier to customize later. Aside that it should be good to go

brucebinary:
@binarydevteam why did you close this? It's not merged yet

brucebinary:
@binarydevteam since you already split this PR to a few I'm closing this

PR #157, opened 2018-04-04
fix bug where UI doesn't instantly update when comparisons is removed…

brucebinary:
…/price update/added

### DIFF
--- a/src/store/ChartStore.js
+++ b/src/store/ChartStore.js
@@ -256,36 +256,40 @@ class ChartStore {
 
         window.addEventListener('resize', this.resizeScreen.bind(this));
 
-        stxx.append('updateChartData', this.updateComparisons);
+        stxx.append('createDataSet', this.updateComparisons.bind(this));
     }
 
-    updateComparisons = (...args) => {
-        /* createDataSet/updateChartData sends more than ten updates per tick.
-            This is to avoid that.
-            Happens only for line chart because of animation
-        */
-        if (args[2] && !args[2].firstLoop) {return;}
-
+    @action.bound updateComparisons(...args) {
         let stx = this.context.stx;
-        let q = stx.currentQuote();
-        const comparisons = [];
-        if (q) {
-            for (const s in stx.chart.series) {
-                const d = stx.chart.series[s];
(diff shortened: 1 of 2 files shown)


brucebinary:
@aminroosta I have some performance concerns with this, since this PR undoes what @4p00rv did before. I'll perform some performance testing soon, but I think from the code it should be fine

src/store/ChartStore.js
```diff
@@ -256,36 +256,41 @@ class ChartStore {
 
         window.addEventListener('resize', this.resizeScreen.bind(this));
 
-        stxx.append('updateChartData', this.updateComparisons);
+        stxx.append('createDataSet', this.updateComparisons.bind(this));
```
aminroosta: `updateComparisons` is already a bound action,  no need for `.bind(this)` call.

src/store/ChartStore.js
```diff
@@ -256,36 +256,41 @@ class ChartStore {
...
-                    symbolObject: p.symbolObject,
-                });
+        const comparisonSymbolsKeys = Object.keys(stx.chart.series);
+        if (comparisonSymbolsKeys.length !== this.comparisonSymbols.length) {
+            const comparisons = [];
+            let q = stx.currentQuote();
+            if (q) {
+                for (const s of comparisonSymbolsKeys) {
+                    const d = stx.chart.series[s];
+                    const p = d.parameters;
```
aminroosta: @brucebinary please avoid single letter variables ... instead of `p` use `prm` or param`. instead of `s` use `sybl` or `key` ... instead of `d` use `seri`

src/store/ChartStore.js
```diff
@@ -256,36 +256,41 @@ class ChartStore {
...
             this.comparisonSymbols = comparisons;
+            return;
         }
-    };
+
+        // Update the comparison prices:
+        let i = 0;
+        for (const s of comparisonSymbolsKeys) {
+            const comp = this.comparisonSymbols[i];
+            const d = stx.chart.series[s];
```
aminroosta: @brucebinary same for single letter variables.

PR #159, opened 2018-04-05
UX staff

binarydevteam:
@brucebinary the problem is that, in my new branch, still the old staff exist


brucebinary:
@binarydevteam this PR has some outdated parts (still using SVG) I think you have a newer PR for the above changes?

PR #161, opened 2018-04-05
Drawing disappear after refresh.

binarydevteam:
@brucebinary  I separate this from the others, make new branch for my each task .

For now, all `target` return same `symbol` if they return different symbol, we can be able to save multiple chart draw


brucebinary:
@binarydevteam huh. I never realized that the drawings are unique to each symbol, even in the chartiq demo. This would meant that with multiple charts, if they all have the same active symbol, the same drawings will be placed on top for each chart. I don't think this is behavior we want. Need to discuss this with UX team.

Good job on this btw.

PR #163, opened 2018-04-05
Ui market closed

binarydevteam:
@brucebinary  please check this code


src/components/CategoricalDisplay.jsx
```diff
@@ -51,6 +51,7 @@ const CategoricalDisplay = ({
         >
             {renderLeft(item)}
             <div className="right">
+                {(item.dataObject && item.dataObject.exchange_is_open == 0 )?<span className="closed-market">{t.translate("CLOSED")}</span>:''}
```
brucebinary: Ha. I forgot that `CLOSED` will need to be translated. Good thing you changed it to a span.

sass/icons/close/ic-market-closed-line.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="32" height="16" viewBox="0 0 32 16">
```
brucebinary: this can be removed

brucebinary:
@binarydevteam one more change (sry I missed this out earlier). https://github.com/binarydevteam/SmartCharts/blob/4979a350d834ea40888b2ce62a61656dec1cb5f2/src/components/CategoricalDisplay.jsx#L50 should be removed, otherwise the market will still display as disabled. Of course, clicking it will not do anything for now. After we add `When user clicks, there will be no live stream; the stream will only retrieve the history instead`, the `onClick` method will need to be updated accordingly.

PR #169, opened 2018-04-06
Add animations

binarydevteam:
@brucebinary  I try to add this animation to all dialog element, but as I do not know all the chart, maybe there are some parts that missed.


brucebinary:
@binarydevteam It's missing the animation when the menu closes. The menu open animation looks good to me, but ultimately afiq will be the judge of that.

Once you done the menu closing animation just ping me.

binarydevteam:
@brucebinary  I just fix the problem of animation on closing dialog

PR #171, opened 2018-04-09
POC for mobile view

binarydevteam:
@brucebinary  @aminroosta 
The way I am going to do this for mobile UI

### DIFF
--- a/src/store/ChartStore.js
+++ b/src/store/ChartStore.js
@@ -296,10 +296,6 @@ class ChartStore {
 
     setIsMobile(status) {
         this.isMobile = status;
-
-        if (this.isMobile && this.rootNode) {
-            this.rootNode.setAttribute('class', 'smartcharts-mobile');
-        }
     }
 
     processSymbols(symbols) {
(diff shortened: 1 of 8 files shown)


src/store/ChartStore.js
```diff
@@ -293,6 +294,14 @@ class ChartStore {
         }
     }
 
+    setIsMobile(status) {
+        this.isMobile = status;
+
+        if (this.isMobile && this.rootNode) {
+            this.rootNode.setAttribute('class', 'smartcharts-mobile');
+        }
```
aminroosta: Set the className inside [Chart.jsx](https://github.com/binary-com/SmartCharts/blob/dev/src/components/Chart.jsx#L47)

```jsx
<cq-context ref={...} class={isMobile ? 'is-mobile' ? ''}> ... </cq-context>
```

binarydevteam:
@aminroosta  `Comparison.jsx` here is the way we do for changing component on mobile mode

binarydevteam:
@aminroosta @brucebinary I juts fix the conflict for this PR,
Please check and merge , so that I can move forward

PR #172, opened 2018-04-09
Menus are clickable even though it is closed

binarydevteam:
@brucebinary  problem occure because I just specify `z-index` which not totally work, so I also add visibility,


brucebinary:
LGTM :+1:

PR #174, opened 2018-04-09
Display history

brucebinary:


src/store/CategoricalDisplayStore.js
```diff
@@ -92,14 +92,14 @@ export default class CategoricalDisplayStore {
...
         let idx = i - 1;
         let id;
-        do {
+        while (idx >= 0) {
             id = this.filteredItems[idx].categoryId;
             if (this.categoryElements[id] !== null) {
                 break;
             }
             idx--;
-        } while (idx >= 0);
```
aminroosta: It's been a while i've not seen a do while, we should have kept it :haircut_man:

PR #175, opened 2018-04-09
no scrolling to the future.

brucebinary:
In the future, we would want more control over this when placing trade visualizations in the chart, but in chart.binary.com it makes more sense to disable it

### DIFF
--- a/src/store/ChartStore.js
+++ b/src/store/ChartStore.js
@@ -146,6 +146,7 @@ class ChartStore {
     startUI() {
         const stxx = this.stxx;
         stxx.chart.allowScrollPast = false;
+        stxx.chart.allowScrollFuture = false;
         const context = new Context(stxx, this.rootNode);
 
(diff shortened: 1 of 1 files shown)


src/store/ChartStore.js
```diff
@@ -146,6 +146,7 @@ class ChartStore {
     startUI() {
         const stxx = this.stxx;
         stxx.chart.allowScrollPast = false;
+        stxx.chart.allowScrollFuture = false;
```
aminroosta: @brucebinary  I think we can also take this from chart props. from the library user.

brucebinary: should the need arises yes, but I think it's better for the user to set the range instead (I just find out about this today): http://documentation.chartiq.com/tutorial-Understanding%20Chart%20Range.html#main

I can't think of a use case where users would just want to scroll indefinitely to emptiness

PR #176, opened 2018-04-09
remove favorites from drawtools.

brucebinary:
May be temporary, depending when UX team come up with UI for favorites for drawtools. Until then we should not show it.

### DIFF
--- a/src/store/SettingsDialogStore.js
+++ b/src/store/SettingsDialogStore.js
@@ -7,6 +7,7 @@ export default class SettingsDialogStore {
     @observable items = []; // [{id: '', title: '', value: ''}]
     @observable title = '';
     @observable stared = false;
+    @observable isFavoritable = true;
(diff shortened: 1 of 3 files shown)


src/store/SettingsDialogStore.js
```diff
@@ -69,6 +71,7 @@ export default class SettingsDialogStore {
             showTabs: this.showTabs,
             stared: this.stared,
             setOpen: this.setOpen,
+            isFavoritable: this.isFavoritable,
```
aminroosta: @brucebinary `isFavoritable` is a computed, use a mobx computed if you want to be format.
Or just use `isFavoritable: !!this.onStared` in the connect() method.

brucebinary: I like the 2nd suggestion. Just implemented that. Thanks for your feedback

PR #179, opened 2018-04-11
fix  Menus are clickable even though it is closed #172 problem on 

binarydevteam:
@brucebinary I just made new PR
to fix it,


sass/components/_chart-controls.scss
```diff
@@ -71,14 +71,17 @@ cq-context {
                 height: auto;
                 margin-top: 1;
                 box-shadow: 0 0 0.625em 0.025em rgba(0, 0, 0, 0.14);
-                transition: bottom 0.2s cubic-bezier(0.25, 0.46, 0.45, 0.94),
```
brucebinary: it's good, though as I mentioned earlier this code is duplicated (2 other places IIRC). If you can find a way to reuse it that'd be great.

PR #185, opened 2018-04-13
Mobile view

binarydevteam:
@brucebinary 
# working on chart mobile view 
- change style of dialog
- modify categorical-display component to be use in mobile mode
- add title & footer  to categorical-display component
- make all dialog fullscreen on mobile view
- modify the structure of chart.jsx item


src/components/Chart.jsx
```diff
@@ -55,19 +55,15 @@ class Chart extends Component {
...
                                 <AssetInformation />
                                 <ComparisonList />
                             </div>
                         </RenderInsideChart>
+                        <ChartTitle />
                         <ChartControls />
                         <Crosshair />
                         <div className="chartContainer primary"> </div>
                         <Loader />
-                        {!isChartAvailable &&
```
brucebinary: you didn't merge `dev` properly. I'd recommend using something like [Meld](http://meldmerge.org/) or similar to handle conflicts

src/components/Chart.jsx
```diff
@@ -55,19 +55,15 @@ class Chart extends Component {
                         </RenderInsideChart>
                         <RenderInsideChart>
                             <div className="cq-top-ui-widgets">
-                                <ChartTitle />
                                 <AssetInformation />
                                 <ComparisonList />
                             </div>
                         </RenderInsideChart>
+                        <ChartTitle />
```
brucebinary: Also you shouldn't move `<ChartTitle />` out of `<RenderInsideChart>` because indicator panels can be placed on top of the chart

src/store/CategoricalDisplayStore.js
```diff
@@ -39,7 +39,7 @@ export default class CategoricalDisplayStore {
         }
     }
     /* isMobile: fill form the ChartStore */
-    @observable isMobile = false;
+    @computed isMobile = false;
```
brucebinary: should be `@computed isMobile = this.mainStore.chart.isMobile;`

Then you can remove
```js
        if (this.mainStore) {
            this.isMobile = this.mainStore.chart.isMobile;
        }
```

brucebinary: @binarydevteam ah you already fixed it. Good. Just the DrawTools translation strings (refer above) and we are good to go

PR #186, opened 2018-04-16
remove past notifications after timeout

brucebinary:


src/store/NotificationStore.js
```diff
@@ -12,13 +14,27 @@ export default class NotificationStore {
 
     @observable messages = [];
 
-    @action.bound notify(text, type = NotificationStore.TYPE_WARNING) {
+    // Duration is in seconds; set to < 0 if you want the notification to remain indefinitely
+    @action.bound notify(text, type = NotificationStore.TYPE_WARNING, duration = 10) {
```
brucebinary: @ashkanx I've set the notifications to be removed after 10 seconds.

ashkanx: How about I change the symbol in just few seconds? do we remove the notification on symbol change?

src/store/NotificationStore.js
```diff
@@ -12,13 +14,27 @@ export default class NotificationStore {
...
+            setTimeout(() => this.removeById(id), duration * 1000);
+        }
+    }
+
+    removeById(id) {
+        this.messages.map((msg, i) => {
+            if (msg.id === id) {
+                this.remove(i);
+            }
+        });
```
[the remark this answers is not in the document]
brucebinary: More readable code! :D Done.

PR #192, opened 2018-04-17
Chart setting

binarydevteam:


src/components/ChartSetting.jsx
```diff
@@ -0,0 +1,168 @@
...
+    onViewLanguage,
+    isViewLanguage,
+    setLanguage
+}) => {
+    const renderMain = () => {
+        return <div>
+            <div className='title'> {t.translate('Settings')} </div>
+            <div className='body'>
+                <div className="ciq-list ciq-list-setting">
+                    <div className="ciq-list-item">
```
brucebinary: `ciq-list-item` will make the cursor to pointer for the whole row; we just want the icons to look clickable.

src/store/ChartSettingStore.js
```diff
@@ -0,0 +1,77 @@
...
+        window.location.reload();
+    }
+
+
+    @computed get getLanguage() {
+        return this.language ? this.language : 'en';
+    }
+
+    @computed get hasActiveView() {
+        return this.view !== '' ? true : false;
```
brucebinary: `!==` operator already returns a boolean. You could just do `this.view !== ''`?

src/store/ChartSettingStore.js
```diff
@@ -0,0 +1,77 @@
...
+        }));
+    }
+
+
+    @action.bound setLanguage(lng) {
+        this.language = lng;
+
+        this.saveSetting();
+
+        window.location.reload();
```
brucebinary: this will need to be separated out from library code soon. But for now leaving it here would be fine

src/store/ChartSettingStore.js
```diff
@@ -0,0 +1,77 @@
...
+
+        try {
+            let setting_string = CIQ.localStorage.getItem('setting'),
+                setting = JSON.parse(setting_string !== '' ? setting_string : '{}');
+
+            this.language = setting.language;
+            this.position = setting.position;
+            this.theme = setting.theme;
+
+        }catch (e){
```
brucebinary: I'm against empty catch blocks (A quick google will show you why). If you're testing out fine, but we don't want this in production code.

src/store/ChartSettingStore.js
```diff
@@ -0,0 +1,77 @@
...
+
+    @computed get hasActiveView() {
+        return this.view !== '' ? true : false;
+    }
+
+    @action.bound clearView() {
+        this.view = '';
+    }
+
+    @computed get isViewLanguage() {
```
brucebinary: maybe you can use a map instead of a query function for each route? 

```js
routes = {
   language: <div>language list blabla</div>,
   theme: <div>theme display yadda yadda</div>
}
```

src/store/ChartSettingStore.js
```diff
@@ -0,0 +1,77 @@
...
+        return this.view === 'language';
+    }
+
+    @action.bound onViewLanguage() {
```
brucebinary: I just think it's too much effort to create a new `@action` each time we want to add a route. This needs a more generic implementation.

binarydevteam: I want to make `ChartSetting` a full customizing component

src/components/ChartSetting.jsx
```diff
@@ -0,0 +1,114 @@
...
+}) => {
+    const renderMain = () => {
+        let HiddenStyle = {
+            display: 'none'
+        }
+        return <div>
+            <div className='title'> {t.translate('Settings')} </div>
+            <div className='body'>
+                <div className="ciq-list-setting">
+                    <div className="ciq-item" style={HiddenStyle}>
```
aminroosta: Do not use inline-styles

src/components/ChartSetting.jsx
```diff
@@ -0,0 +1,114 @@
...
+                    </div>
+                    <div className="ciq-item" style={HiddenStyle}>
+                        <span className="ciq-icon-text">{t.translate('Theme')}</span>
+                        <div className="ciq-action">
+                            <ThemeDarkIcon 
+                            />
+                            <ThemeLightIcon 
+                            />
+                        </div>
+                    </div>
```
aminroosta: If this is only a place holder for future work, just comment out the code.

brucebinary:
@aminroosta has a point. Settings is not going to be in binary-static. @binarydevteam I think we put this PR on hold first until I separate out the app from the library.

binarydevteam:
@brucebinary @aminroosta  ok, for now I fix the code issues

brucebinary:
Hi @binarydevteam. As of now the library and app has been separated. @aminroosta is currently working on the interface of how users of the library can pick and choose which components to appear in the chart controls, but for the time being bundling together with the library should be fine.

I'd like to request some changes for the language (hope they sound reasonable from UX perspective):

 - I think the whole `ciq-list-item-lng` div in the `Language` setting button should be selectable, not just the `>` icon.
 - When choosing a language, the whole `ciq-list-item` div should trigger the button to be clicked.

Once that's done let's get this merged so QA can test this out.

binarydevteam:
@brucebinary  about the `Language` I think as the two above setting are clickable on their icons, so this language should treat the same

src/components/ChartSetting.jsx
```diff
@@ -0,0 +1,113 @@
...
+}) => {
+    const renderMain = () => {
+        return <div>
+            <div className='title'> {t.translate('Settings')} </div>
+            <div className='body'>
+                <div className="ciq-list-setting">
+                    {/*
+                        other chart setting option which not use for now
+                    <div className="ciq-item">
+                        <span className="ciq-icon-text">{t.translate('Position')}</span>
```
brucebinary: I see you added some strings that need to be translated. You'll need to run `yarn translations` to add them to `messages.pot`. [This is documented in README](https://github.com/binary-com/SmartCharts#translations).

Since we are going to add Position and Theme settings eventually, uncomment these options before running translations and comment them out again.

index.html
```diff
@@ -24,13 +24,38 @@
...
-<script src="./dist/mobx-react.js"></script>
-<script src="./dist/smartcharts.js"></script>
+    <div id="root"></div>
+    <script src="./dist/chartiq.min.js"></script>
+    <script src="./dist/react.js"></script>
+    <script src="./dist/react-dom.js"></script>
+    <script src="./dist/mobx.js"></script>
+    <script src="./dist/mobx-react.js"></script>
+    <script src="./dist/smartcharts.js"></script>
+    <script>
```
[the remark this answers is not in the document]
binarydevteam: done

src/components/ChartSetting.jsx
```diff
@@ -0,0 +1,113 @@
...
+                {t.translate('Language')}
+            </div>
+            <div className='body'>
+                <div className="ciq-list ciq-list-language">
+                    {languages.map( (language,index) => {
+                        return <div className="ciq-list-item" key={index} >
+                            <span onClick={()=> setLanguage(language.key) }>
+                                {language.icon}
+                            </span>
+                            <span className="ciq-icon-text">{t.translate(language.name)}</span>
```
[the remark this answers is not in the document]
binarydevteam: I think It need, because For example, we write `Malaysia` in English but its `مالزی` in Farsi.
I can remove it, if you see it's not necessary. I just try to analyze project as I can.

PR #198, opened 2018-04-18
Refactor by ESLint

MahboobehMohammadi:
There is some changes that's done by ESLint.


brucebinary:
Cool. It's very strange that ESLint doesn't automatically correct and fix on my machine. I never got around that.

PR #200, opened 2018-04-19
Mobile view

binarydevteam:


binarydevteam:
@brucebinary  I think all is done, just someone remove `sass/_ciq_doorbell.sass`, I don't know why?

brucebinary:
@binarydevteam doorbell is a customer feedback service, so I separated it out from the library when prepping it to be deployed to npm

src/index.js
```diff
@@ -42,9 +40,12 @@ class SmartChart extends React.Component {
 
     render() {
         const {children, isMobile} = this.props;
+
+        this.chart.setIsMobile(this.props.isMobile);
```
aminroosta: @shayanypn do not set state in render() method.
That's totally frowned up on in READ 😄

brucebinary:
@binarydevteam remove it. It's no longer part of the library

src/store/ChartStore.js
```diff
@@ -309,7 +309,7 @@ class ChartStore {
      * @param {bool} status if true, measn mobile mode is active
      */
     setIsMobile(status) {
-        this.isMobile = status;
+        this.isMobile = !status ? CIQ.isMobile : status;
```
brucebinary: I think it's better to have `CIQ.isMobile` set by the `charts.binary.com` app. I'll set this when I separate out the app. For the time being discard the changes from `index.js` and ChartStore.js`, and I'll have this merged.

src/index.js
```diff
@@ -26,9 +26,11 @@ class SmartChart extends React.Component {
             activeSymbols = data.active_symbols;
         }
         this.chart.setActiveSymbols(activeSymbols);
+        this.mainStore.chart.onSymbolChange = this.props.onSymbolChange;
+    }
 
+    componentWillMount(){
```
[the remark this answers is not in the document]
binarydevteam: well `componentWillMount` run before `componentDidMount`. I need to initial amount of `isMobile` before the component render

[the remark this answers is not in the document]
brucebinary: Ah I didn't notice `Will` and `Did`.

Anyhow. With `isMobile` passed as a prop we don't need `setIsMobile` anymore.

src/index.js
```diff
@@ -41,10 +38,11 @@ class SmartChart extends React.Component {
     }
 
     render() {
-        const {children, isMobile} = this.props;
+        const {children} = this.props;
+
         return (
             <MobxProvider store={this.mainStore}>
-                <Chart lang="en" isMobile={isMobile}>
```
brucebinary: use this instead

src/index.js
```diff
@@ -41,10 +38,11 @@ class SmartChart extends React.Component {
     }
 
     render() {
-        const {children, isMobile} = this.props;
```
brucebinary: use this instead

src/store/ChartStore.js
```diff
@@ -49,7 +49,7 @@ class ChartStore {
     @observable comparisonSymbols = [];
     @observable categorizedSymbols = [];
     @observable barrierJSX;
-    @observable isMobile;
+    isMobile;
```
brucebinary: remove this

PR #202, opened 2018-04-19
Split app from library

brucebinary:


package.json
```diff
@@ -15,11 +15,12 @@
   ],
   "scripts": {
     "build": "rimraf dist && NODE_ENV='production' webpack --optimize-minimize --progress",
-    "build-travis": "yarn build && node scripts/update-hashes.js",
+    "build:app": "BUILD_MODE='app' yarn build",
```
aminroosta: @brucebinary could you please use https://www.npmjs.com/package/cross-env just to stop yarn from choking when used in windows environment.

I know @MahboobehMohammadi is on windows.

PR #204, opened 2018-04-20
changing border-radius,box-shadow, ...

binarydevteam:


sass/_ciq-components.scss
```diff
@@ -9,11 +9,7 @@ cq-context {
 	overflow: hidden;
 
 	input {
-		-webkit-user-select: text;
-		   -moz-user-select: text;
-			-ms-user-select: text;
-				user-select: text;
-
+		@include user-select(none);
```
brucebinary: this should be `@include user-select(text);`

sass/_variables.scss
```diff
@@ -0,0 +1,164 @@
...
+
+
+
+
+/* ------------------------ *\
+        Mixin		
+\* ------------------------ */
+
+/* --------- Animation : dialog --------- */
+@mixin dialog-transition {
```
brucebinary: you can reuse `@mixin transition` here.

A note: including the same mixin that has no change in arguments a few places can result in a pretty big transpiled `css` file with many duplicate code. To save space, I think it better to use 

```scss
a, b, c {
    // ...a lot of style code.
}
```

index.html
```diff
@@ -6,6 +6,7 @@
     <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1" />
     <meta name="description" content="Analyse the financial markets using beautiful visualisations and unique trading tools, powered by a fast and intuitive interface." />
     <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css" media="screen" />
+    <link rel="stylesheet" type="text/css" href="./css/demo.css" media="screen" />
```
brucebinary: the changes in this file is outdated and should be discarded. Pls check before merging next time

PR #205, opened 2018-04-20
Chart controls

MahboobehMohammadi:
ChartControls style for Desktop has been changed.


sass/components/_chart-controls.scss
```diff
@@ -7,12 +7,16 @@ cq-context {
...
         padding: 0 10px;
         transform: translateX(-50%);
-        height: 32px;
-        width: auto;
+        height: 49px;
+        width: 100%;
+
+        flex-wrap: wrap; /* Optional. only if you want the items to wrap */
+        justify-content: center; /* For horizontal alignment */
+        align-items: center; /* For vertical alignment */
```
brucebinary: `/* For horizontal alignment */` and `/* For vertical alignment */` are unnecessary comments.

src/components/Icons.jsx
```diff
@@ -164,35 +164,55 @@ const Wrapper = WrappedComponent => props => {
...
 export const AddIcon = Wrapper(Add);
+export const ZoomInIcon = WrapIconWithSubtitle(Add);
 export const AddThinIcon = Wrapper(AddThin);
-export const CandleIcon = Wrapper(Candle);
+export const CandleIcon = WrapIconWithSubtitle(Candle);
 export const CloseIcon = Wrapper(Close);
-export const ComparisonIcon = Wrapper(Comparison);
-export const CrosshairIcon = Wrapper(Crosshair);
+export const ComparisonIcon = WrapIconWithSubtitle(Comparison);
+export const CrosshairIcon = WrapIconWithSubtitle(Crosshair);
```
brucebinary: I think creating `WrapIconWithSubtitle` is unnecessary complexity that limits the reusability of the icons. We don't need `tooltip-title` attribute anymore; a simpler solution is to place these tooltips in a `<span>` and selectively hide them when `<cq-context/>` has the class `smartcharts-mobile`.

src/components/Timeperiod.jsx
```diff
@@ -15,9 +15,11 @@ const Timeperiod = ({
             className="ciq-period"
         >
             <Menu.Title>
-                <span tooltip-title={t.translate("Time period")}>
+                <span className="bt-priod" tooltip-title={t.translate("Time period")}>
                     <span className="interval_display">{interval_display}</span>
                     <span className="unit_display">{timeUnit_display}</span>
+                    <br/>
+                    <span>{t.translate("Interval")}</span>
```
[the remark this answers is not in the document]
MahboobehMohammadi: Yes.It exists in new design that afiq send in trello.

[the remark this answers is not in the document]
MahboobehMohammadi: image

src/components/Icons.jsx
MahboobehMohammadi: @brucebinary Yes you are right, it could be done by just style but since the icons are being used other palces too, they will still need the tooltip. right?

MahboobehMohammadi: @brucebinary  Yes you are right, it could be done by just styles but since the icons are being used other palces too like when you open Draw Tools there is Measure and Clear All buttons which need to remain as it is with tooltip and without subtitles. right?

src/components/CategoricalDisplay.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { Component, Fragment } from 'react';
 import {CategoryIconMap, ItemIconMap, SearchIcon,
     SymbolPlaceholderIcon, ActiveOptionsIconMap, FavoriteIcon, CloseIcon } from './Icons.jsx';
+import {stxtap} from '../store/utils';
```
brucebinary: Changes in this PR to `CategoricalDisplay.jsx` are outdated; you'll need to revert this. It's always best to handle merge conflicts manually or with a tool (e.g. 3-way merging)

src/components/Icons.jsx
```diff
@@ -152,19 +152,22 @@ import SmartFX from '../../sass/icons/active-symbols/ic-smartfx-placeholder.svg'
 
 const Wrapper = WrappedComponent => props => {
     let { className, ['tooltip-title']: tooltip, ...p } = props;
-    className = `ic-icon ${className ? className : ''} ${tooltip ? 'tooltip' : ''}`;
+    className = `ic-icon ${className ? className : ''}`;
 
     return (
         <span
             className={className}
             tooltip-title={tooltip}
```
brucebinary: this is no longer used right? Should be safe to remove

src/components/CategoricalDisplay.jsx
MahboobehMohammadi: Sure. will do.

src/components/Icons.jsx
MahboobehMohammadi: Since we are using Icon.jsx other places and some needs to show tooltip( you still wants some icons to show tooltip,right? for instance open draw-tools there are two icons that need tooltip), it should be remove from here but for those icons which still needs to show tooltips we should set this class while we're creating them that i will do if its the case.

brucebinary: I see. Then where it is is fine

MahboobehMohammadi: No we have two types of icons, one with tooltip and one without. I added to tooltip class to those who needs it.

PR #210, opened 2018-04-24
add language to WS connection

binarydevteam:
@brucebinary 
about the nam `_connectionManager`, I set this because the store make problem with `connectionManager` name. I don't know why.
Does this renaming make any problem ?


brucebinary:
> about the nam _connectionManager, I set this because the store make problem with connectionManager name. I don't know why.
> Does this renaming make any problem ?

@shayanbinary  Shouldn't be. But considering my review statement above it shouldn't matter

PR #212, opened 2018-04-25
Mobile view

shayanbinary:


src/components/ChartControls.jsx
```diff
@@ -10,9 +10,11 @@ import DrawTools from './DrawTools.jsx';
 import ChartSetting from './ChartSetting.jsx';
 import Share from './Share.jsx';
 
-const ChartControls = () => (
+const ChartControls = ({
+    isMobile
+}) => (
     <div className="cq-chart-controls">
-        <CrosshairToggle />
+        {isMobile ? '' : <CrosshairToggle />}
```
brucebinary: Just make `isMobile` an `@observable`, and have its value set by props. You can then access the value via `this.mainStore.chart.isMobile`. When `isMobile` is changed, all `jsx` files using it will be updated as well.

src/store/ShareStore.js
```diff
@@ -47,10 +47,8 @@ export default class ShareStore {
             window.location.origin : 'https://charts.binary.com';
         this.shareLink = `${origin}#${encodeURIComponent(json)}`;
 
-
         this.loading = true;
-
-        fetch(`${this.bitlyUrl}?access_token=${this.accessToken}&longUrl=${encodeURIComponent(this.shareLink)}`)
+        fetch(`${this.bitlyUrl}?access_token=${this.accessToken}&longUrl=${this.shareLink}`)
```
brucebinary: you change to use `encodeURIComponent` from twice to once. I assume this is some fix or smth...

src/index.js
```diff
@@ -36,6 +36,11 @@ class SmartChart extends React.Component {
             this.mainStore.chart.onSymbolChange = onSymbolChange;
         }
     }
+    componentWillMount() {
+        if ( this.mainStore && this.mainStore.chart) {
+            this.mainStore.chart.setIsMobile(this.props.isMobile);
```
brucebinary: I don't think we need this `setIsMobile` method. Just set `isMobile` directly. If `isMobile` is `undefined` it will be false anyway.

src/store/ShareStore.js
shayanbinary: it's already `encode` above, by re-encoding, it get damaged

src/index.js
shayanbinary: @brucebinary  I am afraid if `this.mainStore.chart` get `undefined`

brucebinary: @shayanbinary hm. True. But you are only setting `isMobile` in one place. What you can do is change the condition to 

`this.mainStore && this.mainStore.chart && this.props.isMobile !== undefined`

and we won't need `setIsMobile`

brucebinary: actually why would `this.mainStore` and `this.mainStore.chart` be `undefined`? The condition should really just be `this.props.isMobile !== undefined`

brucebinary: In fact you shouldn't need to override `componentWillMount`. Just add

`this.mainStore.chart.isMobile = this.props.isMobile || false;`

to `componentDidMount`.

index.html
```diff
@@ -25,12 +25,12 @@
...
 <body>
 <div id="root"></div>
-<script src="./dist/chartiq.min.js"></script>
+<script src="./dist/chartiq.min.js?8a0ff48b7d68fdbf2f32d047857ccbb1"></script>
 <script src="./dist/react.js"></script>
 <script src="./dist/react-dom.js"></script>
 <script src="./dist/mobx.js"></script>
 <script src="./dist/mobx-react.js"></script>
-<script src="./dist/smartcharts.js"></script>
+<script src="./dist/smartcharts.js?cfd0b6fc49e41e49727ce91e89a86d0a"></script>
```
brucebinary: as mentioned in README, don't push the the hash to git

src/components/ChartControls.jsx
```diff
@@ -10,19 +11,43 @@ import DrawTools from './DrawTools.jsx';
...
+    view,
+    share,
+    timeperiod,
+    chartSetting }) => ({
+        isMobile: chart.isMobile,
+        hasOpenMenu: ( 
+            chartType.menu.open || studies.menu.open ||
+            comparison.menu.open || drawTools.menu.open ||
+            view.menu.open || share.menu.open || timeperiod.menu.open ||
+            chartSetting.menu.open
```
brucebinary: this is just a nitpick, but it'll be nice if you can format this to look prettier/readable.

README.md
```diff
@@ -107,11 +107,20 @@ Webpack determines whether to build an app or library depending on whether an en
 
 #### Deploy to [charts.binary.com](https://charts.binary.com/)
 
-> Note: This is usually not required, since Travis will automatically deploy to [charts.binary.com](https://charts.binary.com/) and [charts.binary.com/beta](https://charts.binary.com/beta/) when `master` and `dev` is updated.
+> Note: This is usually not required, since Travis will automatically deploy to [charts.binary.com](https://charts.binary.com/) and [charts.binary.com/beta](https://charts.binary.com/beta/) when `master` and `dev` is
```
brucebinary: why do you remove `updated.`

index.html
```diff
@@ -5,7 +5,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1" />
     <meta name="description" content="Analyse the financial markets using beautiful visualisations and unique trading tools, powered by a fast and intuitive interface." />
-    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css" media="screen" />
+    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css?13a5c9cd5b9b0ebcc44ce4e67a206aad" media="screen" />
```
brucebinary: as mentioned before ^

PR #217, opened 2018-04-30
adding dark theme to chart

shayanbinary:
@brucebinary  Can you please check this PR,
I want to be sure I am in a right way


brucebinary:
- I think you need to have a `clearStyles` when you set a new theme for the chart to clear the old CSS for canvas. Refer http://documentation.chartiq.com/tutorial-CSS%20Overview.html#main
 - binary-static will be passing all chart settings in `ChartSetting.jsx` as prop. Implement those as well, and have the prop values have precedence over those in local storage (`lang` is already doing this). Assume that if prop setting is passed, user will not use `ChartSetting.jsx`.

brucebinary:
@shayanbinary Also once you added those props, add the documentation in README under #API. 

Keep up the good work! (:

sass/_variables.scss
```diff
@@ -45,6 +46,60 @@ $DEFAULT-border:          #e9e9ed !default;
...
+$control-active:          #4b5794 !default;
+$control-bg-active:       rgba(76, 87, 148, 0.2) !default;
+$control-grid:            #fafafb !default;
+$control-dialog-head:     #FFFFFF !default;
+$control-dialog-body:     #FAFAFA !default;
+$control-title-bg:        #FAFAFA !default;
+
+
+
+/******************************\
```
brucebinary: there's no light/dark theme in variables anymore right?

sass/styles/_menu.scss
```diff
@@ -0,0 +1,180 @@
...
+				@include svg-icon-color('activeColor');
+				.ic-subtitle{
+					font-weight: 500;
+					@include themify($themes) {
+						color: themed('controlTextActive');
+					}
+				}
+			}
+		}
+		// > .ic-icon,
```
brucebinary: if it not used just remove it, otherwise document what uncommenting will do

src/components/Chart.jsx
```diff
@@ -63,9 +63,18 @@ class Chart extends Component {
         const insideHolder = array.filter(c => !/(TradeStart)|(TradeEnd)/.test(c.type.displayName));
         const insideSubHolder = array.filter(c => /(TradeStart)|(TradeEnd)/.test(c.type.displayName));
         const renderTopWidgets = topWidgets || defaultTopWidgets;
+        const contextClassName = () =>{
```
brucebinary: this function can be simplified. In fact you may not even need it with template literals

src/store/ChartSettingStore.js
```diff
@@ -116,10 +120,14 @@ export default class ChartSettingStore {
...
+        this.theme = value ? 'dark' : 'light';
+        this.mainStore.chart.stxx.clearStyles();
+        this.saveSetting();
+    }
 
-
-    @action.bound setView(view) {
-        this.view = view ? view : '';
+    @computed get isDarkTheme() {
+        return this.theme === 'dark';
```
brucebinary: this is just nitpicking but... it's a one line function that's only used in one place. You could just pass this condition to the `<Switch />`

brucebinary: @shayanbinary actually never mind this.

sass/components/_categorical-display.scss
```diff
@@ -1,310 +1,359 @@
...
+        line-height: 3.4em;
+        text-align: left;
+        cursor: pointer;
+        @include transition(background-color 0.3s)
+        @include box-sizing(border-box);
+        &.cq-active-filter {
+            font-weight: 800;
+            @include themify($themes) {
+                color: themed('defaultTextActive');
+                border-right: 2px solid themed('activeColor');
```
aminroosta: @shayanbinary Could you please check if `themed(...)` can be used in this way.
I may be wrong but i think `property: value;` syntax dictates `themed(...)` to be the only value.

If that is the case, the fix is obvious
```
border-right-width: 2px;
border-right-style: solid;
@include themify($themes) {
     border-right-color: themed('activeColor');
}
```

Again, it's just an skepticism, let me know if the above works the way it is!

sass/components/_categorical-display.scss
```diff
@@ -1,310 +1,359 @@
...
+    .cq-item, .cq-active-item {
+        padding-left: 20px;
+        padding-right: 20px;
+        display: flex;
+        height: 40px;
+        align-items: center;
+        justify-content: space-between;
+        line-height: initial;
+        @include themify($themes) {
+            border-bottom: 1px solid themed('defaultBorder');
```
aminroosta: same here.

sass/components/_categorical-display.scss
```diff
@@ -1,310 +1,359 @@
...
-                color: $active-color;
-                font-weight: bold;
-                padding: 0.78em 1em;
-                width: 100%;
-                cursor: pointer;
-                @include box-shadow(0,0,0,#fff);
+            cursor: pointer;
+            @include themify($themes) {
+                color: themed('activeColor');
+                @include box-shadow(0,0,0,themed('defaultMain'));
```
aminroosta: same.

sass/components/_categorical-display.scss
shayanbinary: @aminroosta  this piece of code fell me in love with `sass-loader` plugin. to be sure I make  sample on `http://shayanbinary.binary.sx/theme-test/` please click on `chart setting` and change theme. I use bellow code
` @include themify($themes) {
     border: 2px solid themed('activeColor');
}`

PR #221, opened 2018-05-01
bug fix: In comparison list, the loader icon does not show up while p…

brucebinary:
…rices are unavailable

perhaps not a good idea to bundle the loader with the <AnimatedPrice/> component, but mobx does not update the comparison list when the prices change... so...

### DIFF
--- a/sass/components/_comparison-list.scss
+++ b/sass/components/_comparison-list.scss
@@ -59,8 +59,8 @@
     }
     .cq-comparison-loader {
         display: none;
-        height: $icon-size;
-        width: $icon-size;
+        height: 16px;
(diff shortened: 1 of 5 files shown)


sass/components/_comparison-list.scss
```diff
@@ -59,8 +59,8 @@
     }
     .cq-comparison-loader {
         display: none;
-        height: $icon-size;
-        width: $icon-size;
+        height: 16px;
+        width: 16px;
```
aminroosta: Use variables :-)

brucebinary: ok done

PR #227, opened 2018-05-04
Fix ic-icon style bug,styles should not apply to  chart title icons

MahboobehMohammadi:


brucebinary:
Ok closing this PR as per discussed.

PR #229, opened 2018-05-07
Fix up Library for use with Binary Static

brucebinary:


src/index.js
```diff
@@ -1,5 +1,8 @@
 /* eslint-disable no-new, react/jsx-indent, react/no-danger, react/jsx-indent-props */
 import React from 'react';
+// Import babel-polyfill multiple times
+import 'idempotent-babel-polyfill';
```
aminroosta: why do we need [this](https://github.com/codejamninja/idempotent-babel-polyfill/blob/master/src/index.js)?

a simple `import babel-polyfill` would not wok?

PR #235, opened 2018-05-11
Chart setting

shayanbinary:
Notice I merge this branch with my `sass refactoring`


src/components/ChartSetting.jsx
```diff
@@ -6,7 +6,8 @@ import {
     SettingIcon,
     PositionLeftIcon,
     PositionBottomIcon,
-    ChevronRightIcon,
+    ThemeDarkIcon,
+    ThemeLightIcon,
```
brucebinary: We don't use `ThemeDarkIcon` and `ThemeLightIcon` anymore right? It should be safe to remove from icons.jsx

src/store/ChartSettingStore.js
```diff
@@ -74,6 +74,11 @@ export default class ChartSettingStore {
     @observable language = '';
     @observable position = '';
     @observable theme = '';
+    defaultLanguage = {
```
brucebinary: minor tweak: I think this can be changed to just `defaultLanguage = languages[0]`.

Also `languages` doesn't need to be `@observable`

PR #240, opened 2018-05-14
Add symbol prop

brucebinary:
Also:
 - `changeSymbol` is refactored to a method of `ChartStore` instead of a method monkey patched to the context object.
 - fix missing loader when switching between symbols
 - isMobile should default to true


README.md
```diff
@@ -86,6 +86,8 @@ lang | Sets the language.
 chartControlsWidgets | Render function for chart control widgets. Refer to [Customising Components](#customising-components).
 topWidgets | Render function for top widgets. Refer to [Customising Components](#customising-components).
 theme | Sets the chart theme. themes are (`dark\|light`), and default is `light`.
+symbol | Sets the current active symbol.
+isMobile | Switch between mobile or desktop view. Defaults to `true`.
```
aminroosta: isMobile should default to `false`.

src/components/Chart.jsx
```diff
@@ -68,8 +72,8 @@ class Chart extends Component {
 
         const contextClassName = () => {
             let className = '';
-            className += isMobile ? 'smartcharts-mobile' : '';
-            className += (typeof theme === 'string' ) ? ` smartcharts-${theme}` 
+            className += (isMobile === undefined || isMobile) ? 'smartcharts-mobile' : '';
```
aminroosta: nitpick: if you need a default value for a prop use es6 destructuring default values.

`const { ... isMobile = true, ...} = this.props;`

src/store/ChartStore.js
```diff
@@ -268,6 +245,33 @@ class ChartStore {
         stxx.append('createDataSet', this.updateComparisons);
     }
 
+    @action.bound changeSymbol(symbolObj) {
```
aminroosta: Check against current symbol, do not re-render if it's the same symbol.

src/store/ChartStore.js
```diff
@@ -307,6 +311,15 @@ class ChartStore {
         this.stxx = null;
     }
 
+    @action.bound onReceiveProps(nextProps) {
+        if (!this.context) return;
+
+        const { symbol } = nextProps;
+        if (symbol && this.activeSymbols.length > 0 && this.currentActiveSymbol.symbol !== symbol) {
+            this.changeSymbol(symbol);
```
aminroosta: I can't find `onSymbolChange()` getting fired in `this.changeSymbol(symbol)`.

Are we sending the notification?

src/store/ChartStore.js
```diff
@@ -307,6 +311,15 @@ class ChartStore {
         this.stxx = null;
     }
 
+    @action.bound onReceiveProps(nextProps) {
+        if (!this.context) return;
+
+        const { symbol } = nextProps;
+        if (symbol && this.activeSymbols.length > 0 && this.currentActiveSymbol.symbol !== symbol) {
```
aminroosta: When the context is not yet ready or the activeSymbols are empty, can we keep symbol value and use it to render the chart?

```javascript
+        api.getActiveSymbols().then(({ active_symbols }) => {
+            this.setActiveSymbols(active_symbols);
+            if (symbol) {
+                this.changeSymbol(symbol);
+            } else if (stxx.chart.symbol) {
+                this.currentActiveSymbol = stxx.chart.symbolObject;
+            } else {
+                this.changeSymbol(defaultSymbol); // use is here instead of defaultSymbol
+            }
+        });
```

README.md
brucebinary: When binary-static and binary-bot are using this as a library, they will use it as `isMobile` is `true`. For library users this would be the expected setting, unless they want to use the chart as a full app.

src/store/ChartStore.js
brucebinary: @aminroosta no - it is called by the `symbolChange` event from the chart engine: https://github.com/binary-com/SmartCharts/blob/dev/src/store/ChartStore.js#L251

`changeSymbol` fires this event when it creates a new chart.

src/store/ChartStore.js
brucebinary: This is not possible because when we set the current symbol, we need the name, market, submarket names, and pip size as well. What this means is that the chart will render, but the chart title will be blank.

Unless... instead of having the user just pass the symbol code, we have them pass the symbol object itself.

src/store/ChartStore.js
brucebinary: This check is performed in the dropdown (https://github.com/binary-com/SmartCharts/blob/dev/src/store/ChartTitleStore.js#L34) when the symbols are selected, but adding the check here to me is a good safety measure. I'll add this as well.

PR #243, opened 2018-05-15
update height of chartContainer to make space for chart controls

brucebinary:
for some reason `offsetHeight` of `.ciq-chart` doesn't factor the height of the chart controls in beta. Since I can't figure out why, I just add the calculation manually

### DIFF
--- a/app/index.jsx
+++ b/app/index.jsx
@@ -33,7 +33,7 @@ const getLanguageStorage = function(){
 };
 
 const connectionManager = new ConnectionManager({
-    appId: 1,
+    appId: 12812,
     language: getLanguageStorage(),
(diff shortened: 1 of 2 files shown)


app/index.jsx
```diff
@@ -33,7 +33,7 @@ const getLanguageStorage = function(){
 };
 
 const connectionManager = new ConnectionManager({
-    appId: 1,
+    appId: 12812,
     language: getLanguageStorage(),
     endpoint: 'wss://frontend.binaryws.com/websockets/v3',
```
ashkanx: Change WS URL to `wss://ws.binaryws.com/websockets/v3`

brucebinary: @ashkanx done.

PR #250, opened 2018-05-17
bug fix: Chart type doesn't sync with chart upon refresh

brucebinary:
context needs to be resolved *after* the layout data is made available, because the chart type is accessed from the layout data after it is processed by the chart. Because layout data is only made available *after* the active symbols is available (a non-blocking call), it should be placed inside `api.getActiveSymbols().then(({ active_symbols }) => {`

Extra Stuff:
fixing share link problem - @shayanbinary  …
This fix was removed in a past merge conflict (2352a06) so I'm adding it back.


shayanbinary:
@brucebinary  all is good.

PR #251, opened 2018-05-17
Request forget api update

brucebinary:


webpack.config.js
```diff
@@ -27,7 +27,7 @@ const config = {
                     MiniCssExtractPlugin.loader,
                     {
                         loader: 'css-loader',
-                        options: { sourceMap: true, minimize: true }
+                        options: { sourceMap: true }
```
brucebinary: @aminroosta note that `minimize` will be set to `true` when webpack is in production mode

PR #252, opened 2018-05-17
 bug fix: pip size on chart not same with pip size

brucebinary:


shayanbinary:
@brucebinary  all great

PR #274, opened 2018-05-23
Add Candlestick remaining time feature

MahboobehMohammadi:


src/store/TimeperiodStore.js
```diff
@@ -14,11 +15,56 @@ export default class TimeperiodStore {
...
+            var dataSet = stx.chart.dataSet;
+            if(dataSet && dataSet.length != 0 ){
+                let diff = new Date() - dataSet[dataSet.length-1].DT;
+                this.remain = displayMilliseconds((getIntervalInSeconds(stx.layout) * 1000) - diff);
+                this.remain = this.remain ? this.remain : "00:00";
+            }
+        }, 1000);
+    }
+
+    showCandleCountdown = () => {
```
brucebinary: notice that there is a delay before the countdown shows up. This can be fixed by adding `stx.draw()` at the end of this method.

MahboobehMohammadi: @brucebinary what method? you mean setInterval?

src/store/TimeperiodStore.js
```diff
@@ -19,13 +19,59 @@ export default class TimeperiodStore {
...
+            if (totalSec <= 0) { return null; }
+            const padNum = (n) => (`0${n}`).slice(-2);
+            const seconds = padNum(Math.trunc((totalSec) % 60));
+            const minutes = padNum(Math.trunc((totalSec/60) % 60));
+            let hours = Math.trunc((totalSec / 3600) % 24);
+            hours = hours ? `${hours}:` : '';
+            return `${hours}${minutes}:${seconds}`;
+        };
+
+        setInterval(() => {
```
brucebinary: sorry I didn't notice this earlier - this function will be called when the context is loaded. This means that even when the countdown is disabled it will still be called.

src/store/TimeperiodStore.js
```diff
@@ -19,13 +19,59 @@ export default class TimeperiodStore {
...
+            var dataSet = stx.chart.dataSet;
+            if(dataSet && dataSet.length != 0 ){
+                let diff = new Date() - dataSet[dataSet.length-1].DT;
+                this.remain = displayMilliseconds((getIntervalInSeconds(stx.layout) * 1000) - diff);
+                this.remain = this.remain ? this.remain : "00:00";
+                this.showCandleCountdown();
+            }
+        }, 1000);
```
brucebinary: I'm not sure if we should update more often than 1 second. There might be a couple millisecond delay or smth. Maybe we could make it 500 milliseconds instead? If 1 second works fine then sure

src/store/TimeperiodStore.js
MahboobehMohammadi: Done!

src/store/TimeperiodStore.js
MahboobehMohammadi: Done!

src/store/TimeperiodStore.js
MahboobehMohammadi: I think we'd better keep it to 1 second, i see better result with 1 sec interval and i have a slow internet :)  plus when i make it 500 milliseconds, it reduce the time for 2 sec each time like its 10:58 to 10:56, i dont why exactly that happen maybe calling it too times cause a delay, but we'd better stay with 1 sec.

brucebinary: aite then. :+1:

src/store/TimeperiodStore.js
```diff
@@ -119,7 +119,9 @@ export default class TimeperiodStore {
         var stx = this.context.stx;
         var dataSet = stx.chart.dataSet;
         var price = dataSet[dataSet.length-1].Close;
-        return stx.pixelFromPrice(price, stx.chart.panel) + 18;
+        var x = stx.pixelFromPrice(price, stx.chart.panel);
+        x = x > stx.chart.panel.bottom - 60 ? x - 18 : x + 18;
```
brucebinary: it's a nitpick, but I'd prefer you use `let` and `const` over `var` whenever possible.

Also (this one is not a nitpick) do not to use magic numbers (numbers that appear without explanation). Either add a comment explaining or use constants:

```js
x += (x > stx.chart.panel.bottom - WHAT_IS_THIS) ? -PROPER_NAME : +PROPER_NAME ;
```

PR #277, opened 2018-05-24
bug fix: SmartCharts stuck in loading screen in binary-static

brucebinary:
To replicate, set the URL to end with `#/` (e.g. `http://localhost:8080/index.html#/`)

additional bug fix: shared URL should link to charts.binary.com


brucebinary:
@aminroosta how will the implementation work if it is used by library users? Note that binary-static developers use `*.binary.sx` as well.

aminroosta:
@brucebinary  in that case let's make it a config option 

`<SmartCharts  shareEndpoint={"https://charts.binary.com"}  ... />`

binary-static will pass "charts.binary.com" but applications will pass their own href
For example if i want to make it work on my .sx domain all i need to do is

`<SmartCharts shareEndpoint={window.location.href.split('?')[0]} ... />`

aminroosta:
make the `shareEndpoint` optional, defaulting to `https://charts.binary.com`

brucebinary:
@aminroosta ah that works better. But maybe call it `shareOrigin` instead?

aminroosta:
Yea, sure ... sounds good to me

brucebinary:
@aminroosta I just added the prop.

aminroosta:
@brucebinary Let's get this merged.

PR #284, opened 2018-05-25
Crosshair

shayanbinary:


src/store/CrosshairStore.js
```diff
@@ -380,16 +376,15 @@ class CrosshairStore {
...
-        this.crosshair = this.stx.layout.crosshair;
+        this.state = this.stx.layout.crosshair;
         this.init();
     };
 
-    @observable crosshair = false;
-    @action.bound setCrosshair(value) {
-        this.stx.layout.crosshair = value;
+    @action.bound toggleState() {
+        this.state = (this.state == 2) ? 0 : (this.state+1);
```
brucebinary: this can simplified to `this.state = (this.state + 1) % 3`

src/components/CrosshairToggle.jsx
```diff
@@ -1,29 +1,36 @@
...
             <Toggle
-                active={crosshair}
-                onChange={toggleCrosshair}
+                active={state != 0}
+                onChange={toggleState}
                 >
-                <CrosshairIcon
-                    className='ic-icon-with-sub'
-                    tooltip-title={t.translate('Crosshair')} />
+                {state == 0 ? <CrosshairOffIcon className='ic-icon-with-sub'
```
brucebinary: I'm not a big fan of nested ternary operators; they are pretty hard to read. A possible solution is to use an array:

```js
const CrosshairIcon = [CrosshairOffIcon, CrosshairOnIcon, CrosshairTooltipIcon][state];
// ...then use as so:
<CrosshairIcon
    className='ic-icon-with-sub'
    tooltip-title={t.translate('Crosshair')}
/> 

```

Also with this way you reduce duplicate code.

src/components/CrosshairToggle.jsx
```diff
@@ -1,29 +1,36 @@
 import React, { Component } from 'react';
 import {connect} from '../store/Connect';
-import {CrosshairIcon} from './Icons.jsx';
+import {CrosshairIcon, CrosshairOffIcon, CrosshairOnIcon, CrosshairTooltipIcon} from './Icons.jsx';
```
brucebinary: Is `CrosshairIcon` used anymore? If not remove it from here and also remove it from `Icons.jsx`

src/components/CrosshairToggle.jsx
shayanbinary: amazing solution, thank you

src/store/CrosshairStore.js
```diff
@@ -380,16 +376,16 @@ class CrosshairStore {
     };
 
     onContextReady = () => {
-        this.crosshair = this.stx.layout.crosshair;
+        this.state = this.stx.layout.crosshair;
```
brucebinary: IIRC mutating the `@observable` state could immediately trigger mobx to fire a change event. Better to store it in a temporary variable until it is confirmed to change:

```js
const state = this.stx.layout.crosshair;
this.state = (typeof state !== 'number') ? 0 : state;
```

PR #290, opened 2018-05-28
Categorical display ux tweaks

brucebinary:
- In mobile view, the size of the dropdowns will not resize according to the chart height.
 - For stores that are initialised in other stores: Instead of passing in the context, this PR proposes to pass in the `mainStore` instead; it's simpler and has more data that we want (e.g. chart height).
 - Unmounts the Dropdowns when not used. This saves on unnecessary renders.
 - There are some tweaks on the dropdowns that makes the animation smoother.


src/components/Dialog.jsx
```diff
@@ -2,19 +2,19 @@
...
     className,
     isMobile,
     isFullscreen,
+    chartContainerHeight,
+    chartHeight,
 }) => {
     const inner_style = (isMobile && open) ? {
         // reduce chart control panel height except in full screen mode
-        height: `${window.innerHeight - (isFullscreen ? 0 : chartControlHeight)}px`,
-        width: `${window.innerWidth}px`,
```
brucebinary: @shayanbinary do you have a reason for adding this? It seemed to work fine without it.

PR #301, opened 2018-06-04
ChartTitle.update is the root of all evil!

brucebinary:


src/store/ChartStore.js
```diff
@@ -269,7 +269,7 @@ class ChartStore {
         this.resizeObserver = new ResizeObserver(() => this.resizeScreen());
         this.resizeObserver.observe(rootNode);
 
-        stxx.append('createDataSet', this.updateComparisons);
+        this.feed.onComparisonDataUpdate(this.updateComparisons.bind(this));
```
aminroosta: @brucebinary can we remove `.bind(this)` since updateComparisons already has `@action.bound` ?

src/store/ChartTitleStore.js
```diff
@@ -39,52 +38,32 @@ export default class ChartTitleStore {
     }
 
     onContextReady = () => {
-        this.context.stx.append('createDataSet', this.update.bind(this));
+        this.mainStore.chart.feed.onMasterDataUpdate(this.update.bind(this));
         this.update();
     };
 
-    update() {
+    @action.bound async update(quote) {
```
aminroosta: why is this marked as `async` ?

src/store/ChartTitleStore.js
```diff
@@ -39,52 +38,32 @@ export default class ChartTitleStore {
     }
 
     onContextReady = () => {
-        this.context.stx.append('createDataSet', this.update.bind(this));
+        this.mainStore.chart.feed.onMasterDataUpdate(this.update.bind(this));
```
aminroosta: same here ... don't use bind

src/store/ChartTitleStore.js
brucebinary: ops. Was testing out if it would make it faster. It doesn't. Lemme remove it.

src/store/ChartStore.js
brucebinary: Ah. Got it.

PR #303, opened 2018-06-04
Update closed market

shayanbinary:
creating a loop to call `getTradingTimes` each `10 minutes` and update the market closed status


src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
-                } else if (stxx.chart.symbol) {
-                    this.currentActiveSymbol = stxx.chart.symbolObject;
-                    stxx.chart.yAxis.decimalPlaces = stxx.chart.symbolObject.decimal_places;
-                    this.categorizedSymbols = this.categorizeActiveSymbols();
-                    if (onSymbolChange) { onSymbolChange(this.currentActiveSymbol); }
-                } else {
-                    this.changeSymbol(this.defaultSymbol);
-                }
-
+                const today_utc = Date.UTC(
```
brucebinary: You could just use `new Date().getTime()`.

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
+                        (new Date()).getMinutes(),
+                        (new Date()).getSeconds(),
+                    ),
+                    timeToEpochGMT = (time_stirng) => {
+                        let today = new Date(),
+                            time = time_stirng.split(':');
+                        return Date.UTC(today.getUTCFullYear(), today.getUTCMonth(), today.getUTCDate(), time[0], time[1], time[2]);
+                    },
+                    updateActiveSymbols = () => {
+                        api.getTradingTimes().then((data) => {
```
brucebinary: Please find a way to save the result of `api.getTradingTimes()` instead of calling it every minute. This data only changes when a new day comes.

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
+                            if (initialSymbol && !(layoutData && layoutData.symbols)) {
+                                this.changeSymbol(initialSymbol);
+                            } else if (stxx.chart.symbol) {
+                                this.currentActiveSymbol = stxx.chart.symbolObject;
+                                stxx.chart.yAxis.decimalPlaces = stxx.chart.symbolObject.decimal_places;
+                                this.categorizedSymbols = this.categorizeActiveSymbols();
+                                if (onSymbolChange) { onSymbolChange(this.currentActiveSymbol); }
+                            } else {
+                                this.changeSymbol(this.defaultSymbol);
+                            }
```
brucebinary: L276 to L285 should not be inside `updateActiveSymbols()`. You only want to call this once when the chart initialise, not every interval when we refresh the trading times

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
+                        });
+                    };
                 this.context = context;
                 this.contextPromise.resolve(this.context);
                 this.resizeScreen();
                 this.chartPanelTop = holderStyle.top;
+
+                updateActiveSymbols();
```
brucebinary: this line should not be necessary. The `active_symbols` api call already contains data on whether the market is currently opened or closed. Instead of creating a new attribute `is_open` you could simply reuse `exchange_is_open`.

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
-
+                const today_utc = Date.UTC(
+                        (new Date()).getUTCFullYear(),
+                        (new Date()).getUTCMonth(),
+                        (new Date()).getUTCDate(),
+                        (new Date()).getHours(),
+                        (new Date()).getMinutes(),
+                        (new Date()).getSeconds(),
+                    ),
+                    timeToEpochGMT = (time_stirng) => {
```
brucebinary: This can be simplified to:
```js
const timeToEpochGMT = (time_stirng) => {
    const dateStr = new Date().toISOString().substring(0, 11);
    return new Date(`${dateStr}${time_stirng}Z`);
};
```

Also that `time_stirng` typo.

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
+                                                today_utc <= timeToEpochGMT(symbol.times.close[i])
+                                                ) {
+                                                    isOpen = true;
+                                                }
+                                            }
+                                            foundSymbol.is_open = isOpen;
+                                        }
+                                    });
+                                });
+                            });
```
brucebinary: The symbols you have modified are not the ones used by the `CategoryDisplay`'s. You need to update `categorizedSymbols`:

```js
this.categorizedSymbols = this.categorizeActiveSymbols();
```

Also `categorizeActiveSymbols()` uses data from `activeSymbols`, so modifying `active_symbols` won't affect anything.

src/store/ChartStore.js
```diff
@@ -231,25 +231,67 @@ class ChartStore {
...
+                        let today = new Date(),
+                            time = time_stirng.split(':');
+                        return Date.UTC(today.getUTCFullYear(), today.getUTCMonth(), today.getUTCDate(), time[0], time[1], time[2]);
+                    },
+                    updateActiveSymbols = () => {
+                        api.getTradingTimes().then((data) => {
+                            data.trading_times.markets.forEach((market) => {
+                                market.submarkets.forEach((submarket) => {
+                                    submarket.symbols.forEach((symbol) => {
+                                        const foundSymbol = active_symbols.find(item => item.symbol === symbol.symbol);
```
brucebinary: This is a pretty expensive operation. Consider that you are searching an array of 100 symbols 100 times every time you update the trading times. I think a better way would be to modify `exchange_is_open ` attribute directly from `categorizedSymbols` (also do it via an `action` for performance reasons).

It is more complex, but the overall structure of `data.trading_times` and `categorizedSymbols` are identical.

src/store/ChartStore.js
shayanbinary: well As @aminroosta  discuss here , I think its all ok, [slack discuss](https://binary-group.slack.com/archives/G0D2CMTTL/p1527670159000385)

src/store/ChartStore.js
shayanbinary: maybe its just return local epoch time, and the `trade time` is in `GMT`

src/store/ChartStore.js
brucebinary: What we were discussing is _refreshing_ the list every 10 minutes. The way you refresh the list is using the result from `getTradingTimes`. My suggestion here is that because the trading times only change every start of the new day, you don't need to make an API call to the server each time you want to refresh the list every 10 minutes.

src/store/ChartStore.js
brucebinary: `getTime` returns the time in `GMT`. [Refer to docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime):

> getTime() always uses UTC for time representation. For example, a client browser in one timezone, getTime() will be the same as a client browser in any other timezone.

src/store/ChartStore.js
```diff
@@ -315,6 +286,53 @@ class ChartStore {
...
+            const dateStr = new Date().toISOString().substring(0, 11);
+            return new Date(`${dateStr}${time_stirng}Z`).getTime();
+        };
+
+        /**
+         * Get tradeTimes if not loaded yet
+         * OR update the active symbols by comapring open time
+         */
+
+        if (this.tradeTimes.length && this.tradeTimesCapture.getDate() === (new Date()).getDate()) {
```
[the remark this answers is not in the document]
brucebinary: @shayanbinary you're right. Misread a bit.

src/store/ChartStore.js
```diff
@@ -315,6 +286,53 @@ class ChartStore {
...
+                                    isOpen = true;
+                                }
+                            }
+                            foundSymbol.exchange_is_open = isOpen;
+                        }
+                    });
+                });
+            });
+
+            this.setActiveSymbols(this.activeSymbols);
```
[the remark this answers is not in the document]
shayanbinary: @brucebinary first of all I update `this.activeSymbols` then I update `this.categorizedSymbols`. think should work fine.

src/store/ChartStore.js
```diff
@@ -250,6 +251,14 @@ class ChartStore {
...
                 this.resizeScreen();
                 this.chartPanelTop = holderStyle.top;
+
+                /**
+                 * Updating market close status each 10 minute
+                 */
+                this.updateMarketClosedStatus(api);
+                setInterval(() => {
+                    this.updateMarketClosedStatus(api);
+                }, 30 * 1000);
```
shayanbinary: for dev test I put 30second delay, I change it with 10 minute

src/store/ChartStore.js
```diff
@@ -274,6 +283,53 @@ class ChartStore {
...
     }
 
+
+    /**
+     * Get tradeTimes if not loaded yet
+     * OR update the active symbols by comapring open time
+     */
+    @action.bound updateMarketClosedStatus(api) {
+        const todayUtc = (new Date()).getTime();
+        const timeToEpochGMT = (hour) => {
```
shayanbinary: I double check your implementation, as we try to compare with GMT, this way get the GMT time

brucebinary: They both get the GMT time. Note the `Z` at the back. Check the outputs:

```js
const timeToEpochGMTBruce = (time_stirng) => {
            const dateStr = new Date().toISOString().substring(0, 11);
            return new Date(`${dateStr}${time_stirng}Z`).getTime();
};

const timeToEpochGMT = (hour) => {
   let today = new Date(),
                time = hour.split(':');
            return Date.UTC(today.getUTCFullYear(), today.getUTCMonth(), today.getUTCDate(), time[0], time[1], time[2]);
         };
const time = "03:45:00";
console.log(timeToEpochGMTBruce(time));
console.log(timeToEpochGMT(time));
```

src/store/ChartStore.js
```diff
@@ -274,6 +283,53 @@ class ChartStore {
...
+            this.tradeTimes.forEach((market) => {
+                market.submarkets.forEach((submarket) => {
+                    submarket.symbols.forEach((symbol) => {
+                        const foundSymbol = this.activeSymbols.find(item => item.symbol === symbol.symbol);
+                        if (foundSymbol) {
+                            let isOpen = false;
+                            for (let i = 0; i <= symbol.times.open.length; i++) {
+                                if (
+                                    symbol.times.open[i] &&
+                            symbol.times.close[i] &&
```
brucebinary: do you need to check this? You are already looping the contents of `open`, and `open.length` should be the same as `close.length` right?

src/store/ChartStore.js
```diff
@@ -274,6 +283,53 @@ class ChartStore {
...
+                            todayUtc >= timeToEpochGMT(symbol.times.open[i]) &&
+                            todayUtc <= timeToEpochGMT(symbol.times.close[i])
+                                ) {
+                                    isOpen = true;
```
brucebinary: if it is already opened you can break out of this for loop

src/store/ChartStore.js
```diff
@@ -274,6 +283,53 @@ class ChartStore {
...
+
+
+        if (this.tradeTimes.length && this.tradeTimesCapture.getDate() === (new Date()).getDate()) {
+            this.tradeTimes.forEach((market) => {
+                market.submarkets.forEach((submarket) => {
+                    submarket.symbols.forEach((symbol) => {
+                        const foundSymbol = this.activeSymbols.find(item => item.symbol === symbol.symbol);
+                        if (foundSymbol) {
+                            let isOpen = false;
+                            for (let i = 0; i <= symbol.times.open.length; i++) {
```
brucebinary: considering how often you access it, you could take it out to make it look cleaner:

```js
const { open, close } = symbol.times;
```

brucebinary: @shayanbinary this line could be simplified to just `for (let i = 0; i <= open.length; i++) {`

brucebinary:
@shayanbinary why did you close this?...

shayanbinary:
@brucebinary sorry bruce , I fix it

ashkanx:
@shayanbinary conflicts

shayanbinary:
@brucebinary  would you please review my last commit. the problem is that *closed label show in the indicators*.

src/feed/Feed.js
```diff
@@ -98,6 +101,12 @@ class Feed {
                 });
                 callback({ quotes: [] });
                 return;
+            } else {
+                response = await this._binaryApi.getTickHistory(dataRequest);
+                this._mainStore.notification.notify({
+                    text: t.translate(errorMessage),
```
brucebinary: This is not necessary; we pass the language to backend. What comes out here should be translated

src/store/ChartStore.js
```diff
@@ -247,7 +248,15 @@ class ChartStore {
                 } else {
                     this.changeSymbol(this.defaultSymbol);
                 }
+
                 this.setLayoutData(context);
+                /**
+                 * Updating market close status each 10 minute
+                 */
+                this.updateMarketClosedStatus(api);
```
brucebinary: if we are going to `api` all over the class, there's no need to pass it around as function argument. Just save it as `this.api`.

src/feed/Feed.js
```diff
@@ -78,9 +78,12 @@ class Feed {
         this._mainStore.notification.removeByCategory('activesymbol');
 
         if (response.error) {
+            const errorMessage = response.error.message;
             const errorCode = response.error.code;
             const tParams = { symbol: symbolObject.name };
-            if (/^(MarketIsClosed|NoRealtimeQuotes)$/.test(errorCode)) {
+            console.log(errorCode);
```
brucebinary: remove these `console.log` code. Remember that this is library code.

src/store/ChartStore.js
```diff
@@ -247,7 +248,15 @@ class ChartStore {
...
+                setInterval(() => {
+                    this.updateMarketClosedStatus(api);
+                }, 10 * 60 * 1000);
```
brucebinary: I think why QA is not seeing the update is because of the update interval. Not entirely unexpected, since we are updating the trade times every 10 minutes. What if QA sets the trade close time as 10:32 or 11:57?

@aminroosta I've brought this up before. Either we update every minute, or use a timer.

src/feed/Feed.js
```diff
@@ -98,6 +101,12 @@ class Feed {
                 });
                 callback({ quotes: [] });
                 return;
+            } else {
+                response = await this._binaryApi.getTickHistory(dataRequest);
```
aminroosta: @shayanbinary I may be wrong but i don't see `response` being used anywhere.

src/store/ChartStore.js
```diff
@@ -128,7 +130,7 @@ class ChartStore {
             isMobile,
             shareOrigin = 'https://charts.binary.com',
         } = props;
-        const api = new BinaryAPI(requestAPI, requestSubscribe, requestForget);
+        this.api = new BinaryAPI(requestAPI, requestSubscribe, requestForget);
```
aminroosta: @shayanbinary declare `api` as a class member variable, just like `stxx` or `id`.

src/store/ChartStore.js
```diff
@@ -247,7 +248,15 @@ class ChartStore {
...
                 }
+
                 this.setLayoutData(context);
+                /**
+                 * Updating market close status each 10 minute
+                 */
+                this.updateMarketClosedStatus();
+                setInterval(() => {
+                    this.updateMarketClosedStatus();
+                }, 10 * 60 * 1000);
```
aminroosta: @shayanbinary nitpick: `setInterval(this.updateMarketClosedStatus, 10 * 60 * 1000)`;

src/feed/Feed.js
shayanbinary: @aminroosta How about the interval that bruce mention above

src/store/ChartStore.js
```diff
@@ -272,6 +281,62 @@ class ChartStore {
...
+                    });
+                });
+            });
+
+            this.categorizedSymbols = this.categorizeActiveSymbols();
+        } else {
+            this.api.getTradingTimes().then((data) => {
+                this.tradeTimesCapture = new Date();
+                this.tradeTimes = data.trading_times.markets;
+                this.updateMarketClosedStatus();
```
aminroosta: @shayanbinary this function is complicated for not good reason, you don't really need `this.tradeTimes` & that check the if clause (`if( ... && this.tradeTimesCapture.getDate() === (new Date()).getDate())`) is unnecessary.

```js
setInterval(() => {
    const trading_times_promise = this.api.getTradingTimes();
    trading_times_promise.then(this.updateMarketClosedStatus)
                                           .catch(data => console.warn(data.error));
}

// and later here
@action.bound updateMarketClosedStatus(tradingTimes) {
    // simplified implementation
}
```

shayanbinary: @aminroosta  well main purpose of this piece of code to get the trading time each `day`, and update the list each `10 minutes`. we just try to prevent unnecessary API call

aminroosta: @shayanbinary True, but you are calling `this.api.getTRadingTimes()` every 10 min.

aminroosta: @shayanbinary what do you mean by "prevent unnecessary API calls"  here? where is the unnecessary call?

shayanbinary: @aminroosta at first I call API and fill the `this.tradeTimes` then in the time that user is date not change with the *last API Call date* (`this.tradeTimesCapture.getDate() === (new Date()).getDate()`) I just update the trade closed status, if date change, the make new api call

aminroosta: @shayanbinary that optimization should go to `api.getTradingTimes()` (in `BinaryAPI.js`).

shayanbinary: @aminroosta  so you mean I do the same optimization as I did here on  the `BinaryAPI.js`. I mean store data there and return to user if user call the api within `1 day` or if it's within 1 day just don't return anything

aminroosta: @shayanbinary for now just call `api.getTradingTimes()`, don't worry about extra requests.

Later we will optimize `getTradingTimes()` to cache promises (not the data).
Something like

```js
   getTradingTimes() {
       const yyyymmdd = new Date().toISOString().substring(0, 10);
        return this.requestAPICached({
            trading_times: yyyymmdd,
        });
    }

    cached = {};
    requestAPICached(req) {
        const key = JSON.stringify(req);
        if(cached[key]) return cached[key];
        const promise = requestAPI(req);
        cached[key] = promise;
        promise.catch(() => { delete cached[key]; });
        return promise;
    }
```

aminroosta: @brucebinary while writing the above comment i realized the following BinaryAPI.js implementation is incorrect.

```js
    async getActiveSymbols() {
        return this.requestAPI({
            active_symbols: 'brief',
            product_type: 'basic',
        });
    }

    async getTradingTimes() {
        return this.requestAPI({
            trading_times: 'today',
        });
    }
```

All those `async` keywords are useless (& probably a syntax error according to es7 rules), because an async function by definition should `await` for something (a promise).

probably the compiler is generous heree, it just ignores such wrong use cases of `async` keyword.

src/store/ChartStore.js
```diff
@@ -284,55 +282,51 @@ class ChartStore {
      * Get tradeTimes if not loaded yet
      * OR update the active symbols by comapring open time
      */
-    @action.bound updateMarketClosedStatus() {
+    onMarketClosedStatus() {
+        this.api.getTradingTimes()
+            .then(this.updateMarketClosedStatus)
+            .catch(data => console.warn(data.error));
```
shayanbinary: @brucebinary I don't know If this line is necessary or should be removed

src/feed/Feed.js
brucebinary: @shayanbinary when the error code is `NoRealtimeQuotes` we also want to get the tick history. _Other than that_, we don't want to be retrieving the tick history; just display the error message.

src/feed/Feed.js
```diff
@@ -135,7 +144,6 @@ class Feed {
                 });
                 result.quotes = TickHistoryFormatter.formatHistory(response);
             } catch (err) {
-                console.error(err);
```
brucebinary: why did you remove this?

src/store/ChartStore.js
brucebinary: @shayanbinary should be fine if you remove it. I don't really have any preference about it

src/feed/Feed.js
shayanbinary: @brucebinary I just think that is useless. I put it back.

src/feed/Feed.js
shayanbinary: @brucebinary as `NoRealtimeQuotes` need both history and notification, I separate its condition and for the rest of errors just show a notification.

src/feed/Feed.js
```diff
@@ -98,6 +99,19 @@ class Feed {
                 });
                 callback({ quotes: [] });
                 return;
+            } if (errorCode === 'NoRealtimeQuotes') {
```
brucebinary: this could be simplified. Try to cut down the amount of duplicate code in ur PRs

brucebinary:
@shayanbinary could you fix the merge conflicts. It's cause by ur previous PR, so should be easy 4 u to figure out how to fix it.

PR #317, opened 2018-06-08
Separate html2canvas

brucebinary:
Updated library size:

```
              Asset      Size  Chunks                    Chunk Names
    smartcharts.css   141 KiB       0  [emitted]         main
     smartcharts.js   584 KiB       0  [emitted]  [big]  main
smartcharts.css.map   186 KiB       0  [emitted]         main
 smartcharts.js.map  1.02 MiB       0  [emitted]         main
     chartiq.min.js   718 KiB          [emitted]  [big]  
```


src/store/ShareStore.js
```diff
@@ -1,7 +1,10 @@
-import html2canvas from 'html2canvas';
 import { observable, action, reaction, computed, when } from 'mobx';
 import MenuStore from './MenuStore';
 import { downloadFileInBrowser, findAncestor } from './utils';
+import { loadScript } from '../utils';
+
+const html2canvasCDN = 'https://html2canvas.hertzen.com/dist/html2canvas.min.js';
```
brucebinary: I couldn't find a CDN that has the latest version of html2canvas. Older versions don't render all the icons in chart controls.

brucebinary: Wait actually nvm that. Dev ops have us covered. I’ll update the cdn link next week

src/store/ShareStore.js
```diff
@@ -1,7 +1,10 @@
...
+setTimeout(() => loadScript(html2canvasCDN), 2000);
```
aminroosta: I think we should load the scrip on demand, the first time that user clicks download as PNG.

src/store/ShareStore.js
```diff
@@ -120,7 +123,7 @@ export default class ShareStore {
     @action.bound downloadPNG() {
         this.menu.setOpen(false);
         const root = findAncestor(this.stx.container, 'ciq-chart-area');
-        html2canvas(root).then((canvas) => {
+        html2canvas(root).then((canvas) => { // eslint-disable-line no-undef
```
aminroosta: it could look like this,

```javascript
loadHtml2Canvas() {
     if(this._promise_html2canas) {
             return this._promise_html2canvas;
      }
      this._promise_html2canvas = new PendingPromise();
      CIQ.loadScript(html2canvasCDN, () => this._promise_html2canvas.resolve());
}
// later use it in 
@action.bound downloadPNG() {
     // ....
     loadHtml2Canvas()
       .then(() => window.html2canvas(root))
       .then((canvas) => { // ...
       })
}
```

src/store/ShareStore.js
brucebinary: [I initially did that](https://github.com/binary-com/SmartCharts/pull/317/commits/74d16827785031cacdc8ada9be71061a6c01d32c). But then after discussing with mohammed we figured we should have it load after the chart loads, otherwise there will be a slight pause to download a 36Kb file.

src/store/ShareStore.js
brucebinary: do we really need a promise for that? Looks more like an extra step

aminroosta: @brucebinary well, without a promise if you open and close the menu quickly ... you can end up loading the library twice.

have a look at `loadScript` implementation.

brucebinary: ok just added that in.

PR #318, opened 2018-06-10
Enforceactions mobx

shayanbinary:


src/store/ShareStore.js
```diff
@@ -64,17 +75,17 @@ export default class ShareStore {
...
+                    self.setUrlGenerated(false);
                 }
-                self.loading = false;
+                self.setLoading(false);
             })
             .catch(() => {
-                self.loading = false;
-                self.urlGenerated = false;
+                self.setLoading(false);
+                self.setUrlGenerated(false);
```
brucebinary: the point of using actions was to batch the mutations and have the render execute only when all mutations within an action is complete. Doing this way is not that much different from mutating each observable separately. TLDR if there are multiple mutations in one batch, group them to one action.

brucebinary:
@shayanbinary is this PR still used? If not just close it

PR #319, opened 2018-06-10
 enable search by symbol

shayanbinary:


src/store/CategoricalDisplayStore.js
```diff
@@ -184,7 +184,7 @@ export default class CategoricalDisplayStore {
 
         const reg = RegExp(this.filterText, 'i');
         const filterCategory = (c) => {
-            c.data = c.data.filter(item => reg.test(item.display));
+            c.data = c.data.filter(item => reg.test(item.display) || (item.dataObject && reg.test(item.dataObject.symbol)));
```
brucebinary: not exactly an elegant solution, considering `CategoricalDisplayStore` is suppose to be a generic implementation reused in 3 different places with 2 different datasets. But refactoring it would be really tedious...

PR #323, opened 2018-06-12
bring back barriers

brucebinary:


src/components/Barrier.jsx
```diff
@@ -47,16 +48,18 @@ export default connect(
...
-        if (high) { store.high_barrier = high; }
-        if (low) { store.low_barrier = low; }
-        if (relative) { store.relative = relative; }
-        if (draggable) { store.draggable = draggable; }
+        if (high !== undefined) { store.high_barrier = high; }
+        if (low !== undefined) { store.low_barrier = low; }
+        if (relative !== undefined) { store.relative = relative; }
+        if (draggable !== undefined) { store.draggable = draggable; }
         if (onBarrierChange) { store.onBarrierChange = onBarrierChange; }
+        store.setHidePriceLines(hidePriceLines);
```
aminroosta: @brucebinary use `store.hidePriceLines = hidePriceLines;` instead.

The callback the receives `(store, { color, shade ... highPriceLines}) => { }` is already wrapped inside a mobx action in the `connect()` method.

brucebinary: Interesting...

done.

PR #324, opened 2018-06-12
enforce action mobx

shayanbinary:


src/store/ShareStore.js
```diff
@@ -67,19 +64,23 @@ export default class ShareStore {
         hashPromise
             .then((hash) => {
```
brucebinary: How about...

```js
hasPromise
    .then(this.onHashAvailable)
    .catch(this.onHashError);
```

I'm not sure whether you can use [`runInAction`](https://mobx.js.org/refguide/action.html#-runinaction-name-thunk) but I think this is pretty readable to me.

src/store/TimeperiodStore.js
```diff
@@ -78,16 +58,36 @@ export default class TimeperiodStore {
...
             }
 
-            if (callFromSettings) { setRemain(); }
+            if (callFromSettings) { this.setRemain(); }
 
             if (!this.countdownInterval) {
                 this.countdownInterval = setInterval(() => {
-                    setRemain();
+                    this.setRemain();
                 }, 1000);
```
brucebinary: not exactly related to this PR, but now that I do notice it...

`this.remain` should not be an `@obervable`.

src/store/ShareStore.js
shayanbinary: @brucebinary amazing solution

src/store/TimeperiodStore.js
shayanbinary: @brucebinary  I checkout the changes for this file and just remove `@obervable` from `remain`

src/store/ShareStore.js
```diff
@@ -39,14 +39,10 @@ export default class ShareStore {
...
-
     @observable shareLink = '';
 
-    fixedEncodeURIComponent(str) {
-        return encodeURIComponent(str).replace(/[!'()*]/g, c => `%${c.charCodeAt(0).toString(16)}`);
-    }
-    refereshShareLink = () => {
-        const self = this;
+
+    @action refereshShareLink = () => {
```
brucebinary: shouldn't this be `@action.bound refreshShareLink(str) {`?

also do fix the typo. Aside that this looks good to me

src/store/ShareStore.js
```diff
@@ -42,7 +42,7 @@ export default class ShareStore {
     @observable shareLink = '';
 
 
-    @action refereshShareLink = () => {
+    @action.bound refereshShareLink() {
```
brucebinary: Bro. It's "refresh" not "referesh".

PR #327, opened 2018-06-14
Even More Style Fixes

shayanbinary:


sass/components/_chart-controls.scss
```diff
@@ -212,3 +212,40 @@
         }
     }
 }
+@media only screen and (min-width: 480px) {
```
brucebinary: > we could use CSS media queries to selectively hide the chart control labels, but turns out the queries only work for screen sizes. An alternative would be to append class names into the DOM when the chart is a certain width (see resizeScreen in ChartStore).

I've mentioned this in the card. We need to consider that the chart width does not necessary equal to the screen width. Consider binary-static (https://zakibinary.binary.sx/br_new-ui/en/app.html)

sass/styles/_ciq-perfect-scrollbar.scss
```diff
@@ -94,6 +94,17 @@
     width: 11px;
 }
 
+.cq-scroll-panel{
```
brucebinary: Why do you need to change this?

src/store/ViewStore.js
```diff
@@ -25,6 +26,10 @@ export default class ViewStore {
     onContextReady = () => {
         const views = createObjectFromLocalStorage('cq-views');
         if (views) { this.views = views; }
+
+        setTimeout(() => {
```
brucebinary: why is this necessary?

sass/components/_chart-controls.scss
shayanbinary: @brucebinary you absolutely right , in this case I should add some class to `cq-context` so that I can recognize width of chart. what id your Idea about doing in this way ?

sass/styles/_ciq-perfect-scrollbar.scss
shayanbinary: @brucebinary On the first design, when the scroll is inactive, it should hide.

src/store/ViewStore.js
shayanbinary: @brucebinary to initial the PerfectScroll after rendering the list. if I don't do this, I won't be able to make scroll on the list

sass/styles/_ciq-perfect-scrollbar.scss
bruceoutdoors: @shayanbinary perfect scrollbar already does this by default

sass/components/_chart-controls.scss
brucebinary: @shayanbinary I'll leave it to you to decide. I think ChartIQ demo has their convention or smth (but we removed it already lol).

sass/styles/chart.scss
```diff
@@ -121,19 +121,19 @@ html, body {
...
 }
 
 .stx-btn-panel > span.stx-ico-close {
-    background-position: -250px -25px;
+    background-position: -252px -28px;
 }
 
 .stx-btn-panel > span.stx-ico-edit {
-    background-position: -300px -25px;
+    background-position: -303px -29px;
```
shayanbinary: little modifications so that Icons place in the center of the button box

sass/components/_chart-controls.scss
```diff
@@ -212,3 +229,57 @@
         }
     }
 }
+.smartcharts-480 {
```
shayanbinary: place these classes instead of media-query, so that chart control change by the size of chart container but not the screen.

sass/styles/_ciq-perfect-scrollbar.scss
shayanbinary: @brucebinary I put this code so that scroll bar just appears when user scroll, by default it gets appear when the user hovers the scroll area.

src/components/Chart.jsx
```diff
@@ -60,12 +60,14 @@ class Chart extends Component {
...
             showCountdown = false,
             chartContainerHeight,
+            containerWidth,
         } = this.props;
 
         const currentLang = lang || ((setting && setting.language) ? setting.language.key : 'en');
         t.setLanguage(currentLang);
         const currentPosition = `cq-chart-control-${(setting && setting.position && !isMobile) ? setting.position : 'bottom'}`;
         const currentMode = `${isMobile ? 'smartcharts-mobile' : ''}`;
+        const currentWidth =  !isMobile ? `smartcharts-${containerWidth}` : '';
```
brucebinary: `currentWidth` is quite a confusing name. Can you come up with smth more descriptive? I'm also thinking it'll be better if you pass `smartcharts-${containerWidth}` as a prop instead of `containerWidth`

sass/components/_chart-controls.scss
brucebinary: this looks good to me :+1:

src/components/Chart.jsx
shayanbinary: @brucebinary As we do same for the theme, I think it's ok to pass only number

src/components/StudyLegend.jsx
```diff
@@ -32,6 +33,7 @@ class StudyLegend extends Component {
                         className={`ic-icon-with-sub ${menuOpen ? 'active' : ''}`}
                         tooltip-title={t.translate('Indicators')}
                     />
+                    {activeStudiesNo ? <span className={`budget ${activeStudiesNo > 9 ? 'x2' : ''}`}>{activeStudiesNo}</span> : ''}
```
brucebinary: the same code is duplicated in `Comparison.jsx`. Consider turning this to a component. 

Also why do you use the name `budget`? It's called a notification badge.

sass/components/_chart-controls.scss
```diff
@@ -212,3 +236,57 @@
...
+            .cq-menu-btn {
+                > .ic-icon{
+                    padding: .35em 0em;
+                }
+                .ic-subtitle{
+                    display: none;
+                }
+            }
+            .budget{
+                left: 28px;
```
brucebinary: do you have to hard code the value of this class for each width configuration? Small changes would easily cause it to be misaligned. If it is positioned relative to the icon you may not need to do this.

src/components/StudyLegend.jsx
shayanbinary: @brucebinary As Afiq mention on the Trello card, width for 1-digit number should be `16px` and for 2-digit should be `24px`. That's why I use `activeStudiesNo > 9 ? 'x2' : ''`

sass/components/_chart-controls.scss
shayanbinary: @brucebinary Becuase of menu structure I can not handle it by `relative position`. also as we have specific `width` for each size, I think it's ok.

brucebinary: @shayanbinary hm... ok. I'll take a look at this someday.

sass/components/_chart-controls.scss
```diff
@@ -26,9 +26,30 @@
             transform: rotate(225deg); /* 180 + 45 */
             display: none;
         }
+        .cq-menu-btn{
+            position: relative;
+        }
         .cq-menu-dropdown {
             height: auto;
         }
+        .notification-badget {
```
brucebinary: @shayanbinary. Pls. I said "notification badge". Not badget.

src/components/StudyLegend.jsx
brucebinary: > the same code is duplicated in Comparison.jsx. Consider turning this to a component.

@shayanbinary I'm not talking about the width of the badge.

sass/components/_chart-controls.scss
shayanbinary: ooops

src/components/StudyLegend.jsx
shayanbinary: @brucebinary so, I don't get you about this

brucebinary: @shayanbinary create a `<NotificationBadge />` component

src/components/NotificationBadge.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+const NotificationBadge = ({
+    badge,
+}) => {
+    if (!badge) return ' ';
```
brucebinary: should be `return null`

src/components/Comparison.jsx
```diff
@@ -21,7 +22,7 @@ const Comparison = ({
                 className={`ic-icon-with-sub ${menuOpen ? 'active' : ''}`}
                 tooltip-title={t.translate('Comparison')}
             />
-            {activeComparisonsNo ? <span className={`notification-badget ${activeComparisonsNo > 9 ? 'x2' : ''}`}>{activeComparisonsNo}</span> : ''}
+            <NotificationBadge badge={activeComparisonsNo} />
```
brucebinary: consider renaming the prop `badge` to `notificationCount` or something more reflective of what it does

shayanbinary: @brucebinary think the `<MenuBadge  />` should be better

brucebinary: @shayanbinary I'm talking about the prop name not the component name. Component name is fine.

shayanbinary: @bruceoutdoors  does component name is ok? I think Notification is the different object in the app.

brucebinary: @shayanbinary no. At least do google on the words "notification badge" and "menu badge" and see what they return.

shayanbinary: @brucebinary  You are right, I keep this in mind.

src/components/StudyLegend.jsx
```diff
@@ -1,38 +1,51 @@
-import React from 'react';
+import React, { Component } from 'react';
 import Menu from './Menu.jsx';
 import CategoricalDisplay from './CategoricalDisplay.jsx';
+import NotificationBadge from './NotificationBadge.jsx';
 import { connect } from '../store/Connect';
 import { IndicatorIcon } from './Icons.jsx';
 
-const StudyLegend = ({
```
brucebinary: Pls be more careful when handling merge conflicts! You can't just go around undoing what other people have done and have them be cool about it. Check my comment on trello and look at https://github.com/binary-com/SmartCharts/pull/383

PR #328, opened 2018-06-15
Fix chart types not getting translated

brucebinary:
we can't place chartTypes and aggregates as fixed arrays because translate needs to be executed _after_ the language is set.

### DIFF
--- a/src/store/ChartTypeStore.js
+++ b/src/store/ChartTypeStore.js
@@ -18,73 +18,77 @@ import {
 } from '../components/Icons.jsx';
 import SettingsDialogStore from './SettingsDialogStore';
 
-export const chartTypes = [
-    { id: 'mountain',      text: t.translate('Line'),           candleOnly: false, icon: LineIcon         },
-    { id: 'line',          text: t.translate('Dot'),            candleOnly: false, icon: DotIcon          },
-    { id: 'colored_line',  text: t.translate('Colored Dot'),    candleOnly: false, icon: DotIcon          },
-    { id: 'spline',        text: t.translate('Spline'),         candleOnly: false, icon: SplineIcon       },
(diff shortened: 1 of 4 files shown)


src/store/ChartTypeStore.js
```diff
@@ -18,73 +18,77 @@ import {
...
-    { id: 'colored_bar',   text: t.translate('OHLC'),           candleOnly: true,  icon: OHLCIcon         },
-    { id: 'hollow_candle', text: t.translate('Hollow Candle'),  candleOnly: true,  icon: HollowCandleIcon },
-    { id: 'heikinashi',    text: t.translate('Heikin Ashi'),    candleOnly: true,  icon: HeikinAshiIcon   },
-    { id: 'kagi',          text: t.translate('Kagi'),           candleOnly: true,  icon: KagiIcon,        settingsOnClick: true },
-    { id: 'linebreak',     text: t.translate('Line Break'),     candleOnly: true,  icon: LineBreakIcon,   settingsOnClick: true },
-    { id: 'renko',         text: t.translate('Renko'),          candleOnly: true,  icon: RenkoIcon,       settingsOnClick: true },
-    { id: 'rangebars',     text: t.translate('Range Bars'),     candleOnly: true,  icon: RangeBarsIcon,   settingsOnClick: true },
-    { id: 'pandf',         text: t.translate('Point & Figure'), candleOnly: true,  icon: PointFigureIcon, settingsOnClick: true },
-];
+export function getChartTypes() {
```
shayanbinary: @brucebinary everything looks great, but just a question, whats the benefit of converting this to a `function` ?

brucebinary: @shayanbinary hm... guess my commit message is not very clear. The problem with production now is the translations in the chart types is getting executed _before_ the language is set. This PR proposes to retrieve the chart types data only when the context is ready.

shayanbinary:
@brucebinary  approved this pull request.

PR #329, opened 2018-06-18
Mobile tapping

shayanbinary:


src/store/DialogStore.js
```diff
@@ -10,17 +11,27 @@ export default class DialogStore {
...
 
-    @action.bound setOpen(val) {
+    setOpen = debounce((val) => {
+        this.openDialog(val);
+    }, 200);
+
+    @action.bound openDialog(val) {
         if (this.open !== val) {
             this.open = val;
             if (this.open) { setTimeout(() => this.register(), 100); } else { this.unregister(); }
```
brucebinary: if you already have `debounce`, you shouldn't need this `setTimeout` here.

src/utils/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-rest-params */
```
brucebinary: try and figure out why eslint is showing this error. You should not use `arguments`.

shayanbinary: @bru no need for this in the new PR commit

src/utils/index.js
```diff
@@ -30,16 +32,154 @@ export function loadScript(scriptName, cb) {
...
 
+function isObject(value) {
+    const type = typeof value;
+    return value != null && (type == 'object' || type == 'function');
+}
 
-export function debounce(callback, wait, context = this) {
-    let timeout = null;
-    let callbackArgs = null;
+export function debounce(func, wait, options) {
```
brucebinary: instead of copy pasting the implementation, just use the npm package: https://www.npmjs.com/package/lodash.debounce

src/store/DialogStore.js
brucebinary: > if you already have debounce, you shouldn't need this `setTimeout` here.

@shayanbinary

src/store/DialogStore.js
```diff
@@ -10,17 +11,27 @@ export default class DialogStore {
...
+        this.openDialog(val);
+    }, 300, { leading: true, trailing: false });
+
+    @action.bound openDialog(val) {
         if (this.open !== val) {
             this.open = val;
             if (this.open) { setTimeout(() => this.register(), 100); } else { this.unregister(); }
+
+            if (val === true) { // close active dialog.
+                if (activeDialog) { activeDialog.closeDialog(); }
```
brucebinary: do you really need a new function for this (`closeDialog` doesn't need to be an `@action` btw)? Please see if you make the implementation as simple possible. Scrutinise every line.

src/utils/index.js
```diff
@@ -28,3 +30,155 @@ export function loadScript(scriptName, cb) {
     const s = document.getElementsByTagName('script')[0];
     s.parentNode.insertBefore(script, s.nextSibling);
 }
+
+function isObject(value) {
+    const type = typeof value;
+    return value != null && (type == 'object' || type == 'function');
+}
+
+export function debounce(func, wait, options) {
```
brucebinary: instead of copy pasting the implementation, just use the npm package: https://www.npmjs.com/package/lodash.debounce

PR #331, opened 2018-06-18
Optimise Crosshair Performance

brucebinary:
Boost `mouseevent` execution time from 38ms to 6ms.

This PR is also necessary prerequisite for markers to work without hinderance, because in current implementation the `Tooltip` class replaces the default `placementFunction`.


src/store/CrosshairStore.js
```diff
@@ -19,122 +20,57 @@ class Tooltip extends CIQ.Marker {
...
         this.hide = hide;
-        this.lastBar = { };
+        this.lastBar = {};
 
         this.className = 'CIQ.Marker.Tooltip';
         this.showChange = false;
         this.showSeries = true;
         this.showStudies = true;
 
-        CIQ.ChartEngine.prototype.append('undisplayCrosshairs', () => {
```
brucebinary: undisplayCrosshairs injection is not required since renderFunction already checks whether crosshairs are rendered

app/app.scss
```diff
@@ -4,6 +4,123 @@
     display: flex;
 }
 
+.peanut {
```
brucebinary: All CSS added here now are temporary; used experiment with markers

src/store/ChartStore.js
```diff
@@ -152,6 +152,56 @@ class ChartStore {
             yTolerance: 999999, // disable vertical scrolling
         });
 
+        // Monkey patch a custom positionMarkers function because the default selects only
+        // the static placementFunction from CrosshairStore
+        stxx.positionMarkers = function () {
```
brucebinary: this should also be temporary hack. Refer to comment on making Crosshair not a `CIQ.Marker`.

src/store/ChartStore.js
```diff
@@ -152,6 +152,56 @@ class ChartStore {
             yTolerance: 999999, // disable vertical scrolling
         });
 
+        // Monkey patch a custom positionMarkers function because the default selects only
+        // the static placementFunction from CrosshairStore; this causes other markers not to get placed.
+        stxx.positionMarkers = function () {
```
brucebinary: this should also be temporary hack. Refer to [comment](https://github.com/binary-com/SmartCharts/pull/331/files#r195997295) on making Crosshair not a CIQ.Marker.

PR #340, opened 2018-06-20
Dont allow smartchart to save duplicate templates

MahboobehMohammadi:


sass/components/_view.scss
```diff
@@ -86,6 +87,50 @@
 			}
 		}
 	}
+	.cq-view-dialog
+	{
+		margin-top: 100;
```
brucebinary: this doesn't do anything

sass/components/_view.scss
```diff
@@ -86,6 +87,50 @@
 			}
 		}
 	}
+	.cq-view-dialog
```
brucebinary: maybe this should be called `.cq-view-prompt`? Dialog doesn't seem that appropriate a name to use.

Also you are missing that tiny downward pointing arrow below the overwrite button.

src/store/DialogStore.js
```diff
@@ -10,16 +10,18 @@ export default class DialogStore {
 
     @observable open = false;
 
-    @action.bound setOpen(val) {
+    @action.bound setOpen(val , closeActiveDialog = true) {
```
brucebinary: Is `closeActiveDialog` just for that overwrite prompt? Doesn't look very appropriate. That prompt should be a dialog by itself.

src/store/ViewStore.js
```diff
@@ -1,11 +1,13 @@
 import { observable, action, when } from 'mobx';
 import { createObjectFromLocalStorage } from '../utils';
 import MenuStore from './MenuStore';
+import DialogStore from './DialogStore';
 
 export default class ViewStore {
     constructor(mainStore) {
         this.mainStore = mainStore;
         this.menu = new MenuStore(mainStore);
+        this.dialog = new DialogStore(mainStore);
```
brucebinary: how about `this.overwritePrompt`?

src/store/ViewStore.js
```diff
@@ -46,13 +48,29 @@ export default class ViewStore {
...
+            this.updateRoute('main');
             const layout = this.stx.exportLayout();
             this.views.push({ name: this.templateName, layout });
             this.updateLocalStorage();
+            this.templateName = '';
+        }
+    }
+
+    @action.bound onOverwrite() {
+        if (this.templateName.length > 0) {
```
brucebinary: it's odd that we need this. User should not be able to change the input of the template or add anymore templates until they have dealt with the prompt.

src/components/Views.jsx
```diff
@@ -40,8 +46,29 @@ const Views = ({
             />
         </Menu.Title>
         <Menu.Body>
+            <Dialog className="cq-dialog cq-view-dialog">
```
brucebinary: Placing this inside `<Menu.Body>` resizes the menu dropdown when the prompt shows up. The dropdown and the dialog needs to be side by side.

sass/components/_view.scss
```diff
@@ -86,6 +87,50 @@
...
+		padding-top: 10px;
+		display: flex;
+		justify-content:center;
+		>.ic-icon
+		{
+			-moz-transition: none;
+			-webkit-transition: none;
+			-ms-transition: none;
+    		-o-transition: color 0 ease-in;
+			transition: none;
```
brucebinary: You don't need to write so much stuff; use the transition mixin defined in `_variables.scss`

src/store/DialogStore.js
MahboobehMohammadi: when you open a dilaog it closes the dialog that is alrady opened, i want to open a dialog iniside another, so i tell it not to close the old one(the parent dialog here) , so this propery is basicly tell to not close the activedilaog.

brucebinary: aite. Fair point. I'm not sure if this meant we need to keep track of all the active dialogs if we have this setting... Just for one feature it's tedious though.

MahboobehMohammadi: no for the dialog you dont want want  keep track off. like confrim dialogs, in this case if the parrent closes the child that is the confrim dialog will be closed ass well. no need to track it.

src/store/DialogStore.js
```diff
@@ -10,16 +10,18 @@ export default class DialogStore {
 
     @observable open = false;
 
-    @action.bound setOpen(val) {
+    @action.bound setOpen(val , isParentDialog = true) {
```
brucebinary: I like this name better (:

translation/messages.pot
```diff
@@ -317,971 +326,977 @@ msgstr ""
 
 # Indicator strings:
 
-msgid "RSI"
+msgid "RSI"
```
brucebinary: Why is there `^M` here...? These are Indicator strings and practically doesn't change. If you don't want to debug the `yarn translation` code (I suspect it's a windows issue), just retain the translation strings you want and discard the rest.

sass/_themes.scss
```diff
@@ -32,6 +32,7 @@ $themes: (
     controlGrid:            #fafafb,
     controlDialogHead:      #FFFFFF,
     controlDialogBody:      #FAFAFA,
+    controlDialogButton:    #2a3052,
```
brucebinary: Another colour just for this button? We need to get our designers to stick to a certain palette :/ 

Can you use `activeColor` and see if Afiq is ok with that? It's same colour as the modal dialog buttons.

src/store/ViewStore.js
```diff
@@ -78,6 +101,9 @@ export default class ViewStore {
     }
 
     inputRef = (ref) => {
-        if (ref) { ref.focus(); }
+        if (ref) {
+            ref.value = this.templateName;
```
brucebinary: you should not need to do this. Consider using a controlled component https://reactjs.org/docs/forms.html#controlled-components

src/store/ViewStore.js
```diff
@@ -78,6 +101,9 @@ export default class ViewStore {
...
+            ref.focus();
```
brucebinary: if `focus` is only called during rerenders, you mind as well use `autoFocus` and eliminate needing to use `ref`.

PR #341, opened 2018-06-21
Better search

shayanbinary:


src/store/CategoricalDisplayStore.js
```diff
@@ -180,9 +180,12 @@ export default class CategoricalDisplayStore {
             return filteredItems;
         }
 
-        const reg = RegExp(this.filterText.trim(), 'i');
+
+        // regex to check all separate words by comma, should exist in the string
+        const reg = RegExp(`.*(${this.filterText.split(' ').filter(x => x !== '').map(text => text.trim()).join(')+.*(')})+`, 'i');
```
shayanbinary: @brucebinary  by considering the binary-static code I change the regex to treat same as binary-static. thats the thing I change in the code to solve the issue

shayanbinary: in this regex I search in the text to be sure it contain all the `input search text` which user type

src/store/CategoricalDisplayStore.js
```diff
@@ -195,7 +201,11 @@ export default class CategoricalDisplayStore {
             }
         }
 
-        return filteredItems;
+        if (searchHasResult) {
+            this.lastFilteredItems = filteredItems;
+        }
+
+        return this.lastFilteredItems;
```
brucebinary: this may be confusing for people reading your code because it looks like you always return the last filtered items (even if you didn't). Try

```js
        if (searchHasResult) {
            this.lastFilteredItems = filteredItems;
            return filteredItems;
        } else {
            return this.lastFilteredItems;
       }
```

src/store/CategoricalDisplayStore.js
```diff
@@ -200,10 +208,10 @@ export default class CategoricalDisplayStore {
...
         }
-        if (searchHasResult) {
-            this.lastFilteredItems = filteredItems;
-            return filteredItems;
-        }
+
+        /** if any data found, then allow updating lastFilteredItems prop by found data */
+        this.lastFilteredItems = searchHasResult ? filteredItems : this.lastFilteredItems;
+
         return this.lastFilteredItems;
```
brucebinary: Find a way to return `filteredItems` instead. As I mentioned before, returning `this.lastFilteredItems` will confuse people even though the code works

src/store/CategoricalDisplayStore.js
```diff
@@ -177,8 +177,16 @@ export default class CategoricalDisplayStore {
             filteredItems.unshift(activeCategory);
         }
 
+        /**
+        * when search bar is empty or it's the first rendering
+        * lastFilteredItems should contain all data
+        */
+        this.lastFilteredItems = !this.lastFilteredItems.length || this.filterText === ''
```
brucebinary: can you spend more time to make this simpler. It should simply communicate that it will return the last available search result if there is no available result.

PR #342, opened 2018-06-21
set a fix height for offsetHeight

shayanbinary:


src/store/ChartStore.js
```diff
@@ -99,7 +99,7 @@ class ChartStore {
 
     updateHeight(position) {
         const panelPosition = position || this.mainStore.chartSetting.position;
-        const offsetHeight = (panelPosition === 'left') ? 0 : this.chartControlsNode.offsetHeight;
+        const offsetHeight = (panelPosition === 'left') ? 0 : 50;
```
brucebinary: I set it to `this.chartControlsNode.offsetHeight` from `50` from a previous commit. I do not think hard coding 50 is a good solution, as the chart controls height changes based on whether it is in mobile or desktop; you can't assume that it would always be 50.

The reason why the bug happened, is because the height of the chart controls is derived from the chart controls DOM element, and when transitioning from left to bottom, before the chart controls updated the DOM, the offsetHeight is retrieved when it is on the left, which is the height of the chart. This effectively makes the height of the chart 0.

shayanbinary:
@brucebinary think the delay works fine. well it just cause a wait till the `position` setting changed then update the chart

bruceoutdoors:
@shayanbinary um... well it works... do a comment to explain why you need the timeout.

PR #343, opened 2018-06-21
remove japan language

shayanbinary:


src/store/ChartSettingStore.js
```diff
@@ -57,10 +57,6 @@ export default class ChartSettingStore {
             key: 'zh_cn',
             name: '简体中文',
             icon: <FlagIcons.Chinese />,
-        }, {
-            key: 'ja',
-            name: '日本語',
-            icon: <FlagIcons.Japan />,
```
brucebinary: You also need to remove the japanese translation as well. Refer to `Translation.js`

PR #347, opened 2018-06-22
bug fix: if indicator panel is above chart, on refresh the chart titl…

brucebinary:
…e will be on the indicator panel

### DIFF
--- a/src/store/ChartStore.js
+++ b/src/store/ChartStore.js
@@ -29,13 +29,14 @@ class ChartStore {
     chartNode = null;
     chartControlsNode = null;
     chartContainerNode = null;
+    holderStyle;
     @observable context = null;
     @observable currentActiveSymbol;
     @observable isChartAvailable = true;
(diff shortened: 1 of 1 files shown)


src/store/ChartStore.js
```diff
@@ -179,15 +180,16 @@ class ChartStore {
...
         stxx.append('deleteHighlighted', this.updateComparisons);
         stxx.addEventListener('layout', () => {
             this.saveLayout();
-            this.setChartPanelTop(holderStyle.top);
+            this.updateChartPanelTop();
         });
         stxx.addEventListener('symbolChange', this.saveLayout.bind(this));
         stxx.addEventListener('drawing', this.saveDrawings.bind(this));
-        // stxx.addEventListener('newChart', () => { });
+        stxx.addEventListener('newChart', this.updateChartPanelTop);
```
brucebinary: main fix is here. When the chart is first initialised, the chart title needs to be positioned properly

PR #348, opened 2018-06-25
Fixdialog

MahboobehMohammadi:


src/components/Menu.jsx
```diff
@@ -3,12 +3,6 @@ import { CSSTransition } from 'react-transition-group';
 import { stxtap } from '../store/utils';
 
 class Menu extends Component {
-    onOverlayClick = (e) => {
```
shayanbinary: @MahboobehMohammadi Why do you remove this part?

src/components/Chart.jsx
```diff
@@ -137,4 +139,15 @@ export default connect(({ chart, drawTools, studies, chartSetting, chartType })
     chartPanelTop: chart.chartPanelTop,
     setting: chartSetting,
     chartContainerHeight: chart.chartContainerHeight,
+     hasOpenMenu: (
```
shayanbinary: @MahboobehMohammadi its not about this issue, but please check by @brucebinary  whether its good to pass `hasOpenMenu` to `<ChartControls hasOpenMenu={hasOpenMenu} .../>` and use it there. because we do the same in that component (prevent duplicate code)

src/components/Menu.jsx
MahboobehMohammadi: @shayanbinary The dialog will be closed on `handleClickOutside` if user click outside of it and also it will be closed when another dialog opens, so this code realy never uses, and we certainly dont want the dialog to close when we click on it unless we tell it to. so this code can be safely  removed.

src/components/Chart.jsx
MahboobehMohammadi: @shayanbinary good point

src/components/Menu.jsx
MahboobehMohammadi: @shayanbinary I just realized this had a use afterall. since $(document).click() not working correctly on iPhone, this event was using to close the open dialog when user click outside it, but since it was causing problems on android devices after i made the dialog to fill the screen, i just fixed that problem some other way. if you wanna know more check this:
https://stackoverflow.com/questions/3705937/document-click-not-working-correctly-on-iphone-jquery

MahboobehMohammadi: They still can!

app/app.scss
```diff
@@ -42,3 +42,8 @@
 #doorbell-background {
     opacity: 0;
 }
+
+.smartcharts-mobile .cq-chart-controls .ciq-menu .cq-menu-overlay
+{
+    height: calc(var(--view-height) - 39px) !important;
```
aminroosta: @MahboobehMohammadi @brucebinary Css variables are still [new](https://caniuse.com/#feat=css-variables), only the latest version of browsers are supporting them.

src/components/Chart.jsx
```diff
@@ -70,10 +71,11 @@ class Chart extends Component {
         const insideSubHolder = array.filter(c => /(TradeStart)|(TradeEnd)/.test(c.type.displayName));
         const renderTopWidgets = topWidgets || defaultTopWidgets;
 
-
         const defaultTheme = (setting && setting.theme) ? setting.theme : 'light';
         const defaultCandleCountdown = (setting && setting.countdown) ? setting.countdown : false;
 
+        document.getElementById('root').style.setProperty('--view-height', window.innerHeight +'px');
```
aminroosta: @MahboobehMohammadi I may be wrong but i think it only works on `/app`, we don't have a `#root` element in binary-static.

aminroosta: And the second thing is the use of css variables, which is a fantastic idea, but still pretty new and only supported by latest browser versions.

src/components/Chart.jsx
```diff
@@ -70,10 +72,11 @@ class Chart extends Component {
         const insideSubHolder = array.filter(c => /(TradeStart)|(TradeEnd)/.test(c.type.displayName));
         const renderTopWidgets = topWidgets || defaultTopWidgets;
 
-
         const defaultTheme = (setting && setting.theme) ? setting.theme : 'light';
         const defaultCandleCountdown = (setting && setting.countdown) ? setting.countdown : false;
 
+        document.getElementById('root').style.setProperty('--view-height', `${window.innerHeight}px`);
```
brucebinary: I thought we are not going to use CSS variables?

sass/styles/main.scss
```diff
@@ -516,6 +516,14 @@ cq-context {
 	font-family: $base-font;
 	font-size: $font-size-base;
 	overflow: hidden;
+	&.cq-dialog-context{
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary You wanted to Dropdowns and Dialogs to  fill the screen on mobile!? nobody said anything about any header!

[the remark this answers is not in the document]
brucebinary: @MahboobehMohammadi I'm not sure you follow. Whatever feature you develop, you need to take to consideration how it will work in binary-static, or even binary-bot. Yes, we do want it to fill the screen, but we only _only_ the dialog to fill the screen, not the chart.

app/index.jsx
```diff
@@ -63,6 +63,17 @@ const requestSubscribe = streamManager.subscribe.bind(streamManager);
 const requestForget = streamManager.forget.bind(streamManager);
 const shareOrigin = window.location.href.split('?')[0];
 
+
+// Update menu height to show chartControl on top of dialog in mobile devices.
+const style = '.smartcharts-mobile .cq-chart-controls .ciq-menu .cq-menu-overlay';
```
brucebinary: do you need to do this? Would just modifying `app.scss` work?

sass/styles/main.scss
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary  i add the `&.cq-dialog-context` class to the chart only when any dialog is open, when the dialog closes i remove this class, so we shouldnt have any problem.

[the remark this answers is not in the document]
brucebinary: @MahboobehMohammadi I'm aware. But the behaviour is still not what we want. Recall again, and overlays are semi-transparent. We expect when we open a dialog, _just_ the dialog opens, nothing else in the background changes. Would you not expect that as well?

app/index.jsx
MahboobehMohammadi: How else i can use javascript in css? I need `window.innerHeight` to calculate the actual vh in ios devices.its why i said we should use props or css variables, otherwise i have to overwrite the style using js.

brucebinary: true that

brucebinary:
Hey all. As discussed, we realised that we have underestimated the difficulty involved in this task. The current PR would not be able to accommodate the use cases our designers want. @aminroosta will be taking over for this task.

Closing this PR now

PR #349, opened 2018-06-25
Barrier bug fixes

brucebinary:
- shade color is not set using `shadeColor`, and only a choice of `green` and `red`
 - you can now set color of barrier lines via `color`.
 - if barrier is not `draggable`, it will rendered differently
 - fix documentation:

   - should be `onChange` not `onBarrierChange`
   - should be `hidePriceLines` not `disablePriceLines`


app/test.jsx
```diff
@@ -131,8 +137,9 @@ class App extends React.Component {
                         &nbsp;
                         <b>low:</b> <input id="low" type="number" value={low === undefined ? '' : low} onChange={this.onHighLowChange} />,
                         <b>high:</b> <input id="high" type="number" value={high === undefined ? '' : high} onChange={this.onHighLowChange} />;
-                        No PriceLine: <input type="checkbox" onChange={this.onPriceLineDisableChange} />
-                        Relative: <input type="checkbox" onChange={this.onRelativeChange} />
+                        No PriceLine: <input type="checkbox" checked={hidePriceLines === undefined ? '' : hidePriceLines} onChange={this.onPriceLineDisableChange} />
```
jbaek7023: Is it better to check your variable with typeof? 
`typeof hidePriceLines != 'undefined'`

This will fail of you have not declared the variable 
`hidePriceLines` with error `Uncaught ReferenceError: params is not defined`

aminroosta:
@brucebinary Please check this with  dark-color, check if `#FFF` for draggable barriers on dark theme looks good.

brucebinary:
@aminroosta it doesn't. But I'll worry about that once the design team comes back with the dark theme design.

aminroosta:
@brucebinary fair enough

app/test.jsx
brucebinary: @jbaek7023 I'm not sure I get your point. It's `typeof t === "undefined"` vs simply `t === undefined`. Why will I get a `ReferenceError`?

brucebinary:
@aminroosta it will work with `rgb(...)` as well. In fact, React converts `#000` to `rgb(0,0,0)` when rendering the DOM.

app/test.jsx
jbaek7023: @brucebinary  It's for the safer usage and nothing else. 👍 

Reference: https://stackoverflow.com/questions/16453569/comparing-undefined-value-in-javascript

brucebinary: if that is the reason I prefer to have that error thrown. I don't see how is it safer. If I'm referencing something that doesn't exist anywhere in context I do not expect it to work. Here, as I am accessing the values from an object, it will always be either `undefined` or something else.

PR #351, opened 2018-06-26
Barrier bug fixes

brucebinary:


src/components/RenderInsideChart.jsx
```diff
@@ -11,18 +13,23 @@ class RenderInsideChart extends PureComponent {
...
-            this.marker = marker;
+            const nodeName = `${inChartPrefix}${at}`;
+            // reuse existing node when possible:
+            let elem = context.topNode.querySelector(`.${nodeName}`);
+            if (!elem) {
+                elem = createElement(`<div class="${nodeName}"></div>`);
+                context.stx.chart.panel[at].appendChild(elem);
+            }
+            this.container = elem;
+            this.forceUpdate(); // force render to be called after getting the container
```
brucebinary: When repeatedly mounting and unmounting `<RenderInsideChart/>`, render is not called after children is `this.container` is set. We need to force render here.

src/components/Barrier.jsx
```diff
@@ -53,19 +56,22 @@ export default connect(
...
-        if (draggable !== undefined) { store.draggable = draggable; }
-        if (onChange) { store.onBarrierChange = onChange; }
-        store.lineStyle = lineStyle;
-        store.hidePriceLines = !!hidePriceLines;
+        store.initializePromise.then(action(() => {
+            if (color) { store.color = color; }
+            if (shadeColor) { store.shadeColor = shadeColor; }
+            if (shade) { store.shadeState = `SHADE_${shade}`.toUpperCase(); }
+            if (isValidProp(high)) { store.high_barrier = high; }
+            if (isValidProp(low)) { store.low_barrier = low; }
```
brucebinary: prevent `NaN`, `null` and `undefined` from being passed

src/components/Barrier.jsx
```diff
@@ -53,19 +56,22 @@ export default connect(
...
-        if (shadeColor) { store.shadeColor = shadeColor; }
-        if (shade) { store.shadeState = `SHADE_${shade}`.toUpperCase(); }
-        if (high !== undefined) { store.high_barrier = high; }
-        if (low !== undefined) { store.low_barrier = low; }
-        if (relative !== undefined) { store.relative = relative; }
-        if (draggable !== undefined) { store.draggable = draggable; }
-        if (onChange) { store.onBarrierChange = onChange; }
-        store.lineStyle = lineStyle;
-        store.hidePriceLines = !!hidePriceLines;
+        store.initializePromise.then(action(() => {
```
brucebinary: this solves the race condition issue in binary-static; no props will be set, nor barrier will be rendered, until the chart has data.

PR #360, opened 2018-07-05
Chart settings refactor

brucebinary:
chart settings can now only be modified via props

reading and writing to local storage, or refreshing the page when language changes, is now the responsibility of the app.

Chart settings API has been changed: https://github.com/brucebinary/SmartCharts/tree/chart-settings-refactor#chart-settings


app/index.jsx
```diff
@@ -67,18 +68,47 @@ const requestSubscribe = streamManager.subscribe.bind(streamManager);
...
+        this.setState({ settings });
+
+        if (this.startingLanguage !== settings.language) {
+            window.location.reload();
+        }
+    };
+
+
+    render() {
+        const { settings } = this.state;
```
brucebinary: SFC doesn't rerender when settings is changed, so I store as a state instead.

src/components/Chart.jsx
```diff
@@ -39,6 +39,11 @@ class Chart extends Component {
         this.props.init(this.root, this.props);
     }
 
+    componentWillReceiveProps(nextProps) {
+        const { settings, setSettings } = nextProps;
+        setSettings(settings);
```
brucebinary: this is where the settings will changed via props

MahboobehMohammadi:
LGTM

PR #363, opened 2018-07-06
Move notification out of app

shayanbinary:


app/index.jsx
```diff
@@ -10,13 +10,17 @@ import { // eslint-disable-line import/no-extraneous-dependencies,import/no-unre
...
+    AssetInformation,
+    ComparisonList,
 } from '@binary-com/smartcharts'; // eslint-disable-line import/no-unresolved
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { configure } from 'mobx';
 import './app.scss';
 import './doorbell';
 import { ConnectionManager, StreamManager } from './connection';
+import Notification from './libs/Notification.jsx';
```
brucebinary: just put it as part of `app`; why do we need a `lib` folder for 1 file?

app/index.jsx
```diff
@@ -61,25 +64,49 @@ const renderControls = () => (
         <ChartSetting />
     </React.Fragment>
 );
-
 const requestAPI = connectionManager.send.bind(connectionManager);
 const requestSubscribe = streamManager.subscribe.bind(streamManager);
 const requestForget = streamManager.forget.bind(streamManager);
 const shareOrigin = window.location.href.split('?')[0];
+function Notifier() {
```
brucebinary: make this a class. Use a separate file, or make it part of Notification component

src/components/Chart.jsx
```diff
@@ -25,6 +25,7 @@ const defaultTopWidgets = () => (
         <ChartTitle />
         <AssetInformation />
         <ComparisonList />
+        <Notification />
```
brucebinary: How could this work? `<Notification />` is in `app`.

src/store/ChartStore.js
```diff
@@ -108,6 +112,14 @@ class ChartStore {
         this.chartContainerHeight = this.chartHeight - offsetHeight;
     }
 
+    notify(message) {
```
brucebinary: couldn't it just be 

```js
notify(message) {
    if (this.onMessage) { this.onMessage(message); }
}
```

app/index.jsx
shayanbinary: @brucebinary  sure, but I have a problem with passing messaging between chart and notification component, I am thinking about using mobx here.

shayanbinary:
@brucebinary there is a problem with the notification position, for now, its appear beside chart title. do you have any Idea how we can change it? (that's because we place it in the `topWidgets` prop)

brucebinary:
@shayanbinary where do you want to place it?

shayanbinary:
@brucebinary Is that meaningful to add another prop like `topWidgets` for notification ?

src/feed/Feed.js
```diff
@@ -73,25 +72,25 @@ class Feed {
         let response = await tickHistoryPromise;
 
         // Clear all notifications related to active symbols
-        this._mainStore.notification.removeByCategory('activesymbol');
+        // this._mainStore.notification.removeByCategory('activesymbol');
```
brucebinary: just remove it. Leaving it will confuse people in the future.

Btw this code was added for a reason. Ashkan wanted all active symbols related notifications to be cleared when the symbol is changed. To do this in `app` subscribe to `onSymbolChanged` event and clear those notifications by category.

src/store/ChartStore.js
```diff
@@ -108,6 +109,10 @@ class ChartStore {
...
+        if (this.messaging) { this.messaging(message); }
```
brucebinary: consider `onMessage` instead of `messaging`. `messaging` implies a status, not an event handler.

app/index.jsx
```diff
@@ -88,16 +106,27 @@ class App extends Component {
...
+    startingLanguage = 'en';
     render() {
+        const renderTopWidgets = () => (
+            <React.Fragment>
+                <ChartTitle />
+                <AssetInformation />
+                <ComparisonList />
+                <Notification
+                    messages={this.state.messages}
+                    onRemove={id => this.onDestroyMessage(id)}
```
brucebinary: The original `<Notification />` component has a timeout before the notification hides itself. If you need to rewrite a component, pls make sure it retains its original functionality

app/index.jsx
```diff
@@ -88,16 +106,27 @@ class App extends Component {
...
-
+    startingLanguage = 'en';
     render() {
+        const renderTopWidgets = () => (
+            <React.Fragment>
+                <ChartTitle />
+                <AssetInformation />
+                <ComparisonList />
+                <Notification
+                    messages={this.state.messages}
```
brucebinary: I see the rationale behind this design decision. You wanted to maintain React's prop style of passing data. But it's also odd to have the logic of managing the notifications passed to the parent. It meant anyone who wants to use your `<Notification />` needs to have some version of `onMessage` and `onDestroyMessage`, wherever they use it.

Unless you have a better solution, I still think the `Notifier` design is better for this case.

app/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+        this.setState({
+            messages,
+        });
+
+        setTimeout(() => {
+            messages = messages.filter(x => x.id !== id);
+            this.setState({
+                messages,
+            });
+        }, 300);
```
aminroosta: @shayanbinary why we need the 300ms delay? what happens if OnMessage gets called with a new message within that 300ms?

app/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+    onRemoveAll() {
+        this.state.messages.forEach((x) => {
+            this.onRemove(x.id);
+        });
```
aminroosta: Just do `this.SetState({ messages: []})`

app/index.jsx
```diff
@@ -62,42 +65,63 @@ const renderControls = () => (
...
+            this.callback({
+                action: _action,
+                data: _data,
+            });
+        }
+    };
+    this.onCallback = (callback) => {
+        this.callback = callback;
+    };
+}
```
aminroosta: Make this a class, rename to ChartNotifier
Move to it's own file.

app/Notification.jsx
shayanbinary: If `OnMessage` get called just new message appear, as you see we assign an id to each message `message.id = (new Date()).getTime();`.

aminroosta: I see, makes sense now :)

app/ChartNotifier.js
```diff
@@ -0,0 +1,17 @@
+
+
+function ChartNotifier() {
+    this.callback = null;
+    this.notify = (_action, _data) => {
+        if (typeof this.callback === 'function') {
```
brucebinary: this line should not be needed.

app/Notification.jsx
brucebinary: > message removing has an animation which animate 300ms so, first of all, I set hide=true to start animation then after the animated finish, I remove the message for the store

That's a good explanation 👍 . I think it should be in the code as well.

app/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    render() {
+        const { notifier } = this.props;
+        notifier.onCallback((e) => {
+            if (e.action === 'message') {
+                this.onMessage(e.data);
+            } else if (e.action === 'removeall') {
```
brucebinary: I'm not very fond of this "removeall" thing. Recall that Ashkan wanted _symbol related_ notifications to be removed, not all notifications. You need a `removeByCategory` method in your `ChartNotifier` class

app/Notification.jsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import './_ciq-notification.scss';
+
+import CloseIcon from './icons/ic-close.svg';
+import Warning from './icons/warning.svg';
+import Error from './icons/error.svg';
+import Success from './icons/success.svg';
+import Info from './icons/info.svg';
```
brucebinary: charts.binary.com is designed to use smartcharts as an app. Adding these this way would fail cause the environment in `app` does not have `inline-react-svg`. Refer to https://github.com/binary-com/SmartCharts#debugging-npm-package

app/ChartNotifier.js
```diff
@@ -0,0 +1,17 @@
...
+            this.callback({
+                action: _action,
+                data: _data,
+            });
+        }
+    };
+    this.onCallback = (callback) => {
```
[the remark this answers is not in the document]
shayanbinary: @brucebinary `notifier.onCallback((e) => {` in the `render` method of `Notification.jsx` just call a method each time a message comes to `ChartNotifier`, I try to make a messaging(store) here, to call a method whenever a message comes.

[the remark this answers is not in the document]
brucebinary: @shayanbinary I see. But you don't have to keep assigning `onCallback` in the render method. Besides, the implementation you put in `render` in `Notification` looks like you could simply place it inside `ChartNotifier`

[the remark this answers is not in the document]
shayanbinary: @brucebinary let me see if I understand you or not.

app/Notification.jsx
shayanbinary: @brucebinary, unfortunately, I can not make this work

app/ChartNotifier.js
```diff
@@ -0,0 +1,22 @@
+
+
+export default class ChartNotifier {
+    messageCallback = null;
+    removeByCategoryCallback=null
+
+    notify(_action, _data) {
+        if (_action === 'message') {
+            this.messageCallback(_data);
+        } else if (_action === 'removeByCategory') {
```
brucebinary: just call `removeByCategory`. Why do you need `notify('removeByCategory', _data)` to call `removeByCategory`?

app/Notification.jsx
```diff
@@ -0,0 +1,85 @@
...
+            });
+        }, 300);
+    }
+    onRemoveAll() {
+        this.state.messages.forEach((x) => {
+            this.onRemove(x.id);
+        });
+    }
+
+    removeByCategory(category) {
```
brucebinary: this should be `removeByCategory({ category }) {`

app/ChartNotifier.js
shayanbinary: I use `notify` for both notifying message and remove messages by category, Should I use separate these two methods?  `notify('removeByCategory', _data)` I use `_data` to pass category.

brucebinary: @shayanbinary  yes it should be 2 methods. Think about it. What does verb `notify` have to do with removing notifications?

brucebinary: @shayanbinary also. README documentation for `onMessage`?

app/ChartNotifier.js
```diff
@@ -0,0 +1,22 @@
...
+
+    removeByCategory(category) {
+        this.removeByCategoryCallback(category);
+    }
+
+    setMessageCallback(callback) {
+        this.messageCallback = callback;
+    }
+
+    setRemoveByCategoryCallback(callback) {
```
brucebinary: @shayanbinary do you think it should be `onMessage` and `onRemoveByCategory`? I think that's the convention in general for assigning event handlers.

app/Notification.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import './_ciq-notification.scss';
+
+class Notification extends React.Component {
+    constructor(props) {
```
brucebinary: I think we can assert that `Notification.jsx` would not work without a `ChartNotifier`, so we can assign the event handlers in the constructor, instead of assigning it in the `render` method.

app/Notification.jsx
```diff
@@ -0,0 +1,82 @@
...
+        super(props);
+        this.state = { messages: [] };
+        props.notifier.onMessage(this.onMessage);
```
brucebinary: just a nitpick: you can use destructuring for this 

    const { notifier } = this.props;

PR #367, opened 2018-07-07
Search and category-select auto hide when scrolled up and down

MahboobehMohammadi:
search and category-select auto hide when scrolled down, but show up when scrolled up in mobile


src/store/CategoricalDisplayStore.js
```diff
@@ -52,8 +52,9 @@ export default class CategoricalDisplayStore {
         emptyDescription: t.translate('There are no favorites yet.'),
         data: [],
     };
+    @observable hideLookupFilter = false;
     scrollTop = undefined;
-
+    autoHideLookupFilter = true;
```
brucebinary: why do we need this variable?

src/components/CategoricalDisplay.jsx
```diff
@@ -102,7 +103,7 @@ const CategoricalDisplay = ({
     return (
         <div className="cq-categorical-display">
             {renderMobileTitle()}
-            <div className="cq-lookup-filters">
+            <div className={`cq-lookup-filters ${hideLookupFilter ? 'hide' : ''}`}>
```
brucebinary: `hide` could be rather misleading, because in desktop this shows up as well. Perhaps `scroll-down` is more appropriate?

src/store/CategoricalDisplayStore.js
MahboobehMohammadi: @brucebinary When user search or moving through categories by selecting them, the scroll moves up and down as well, its inconvenient if we dispaear the seach text or category-select when user are using them. this is tell not to hide them on those cases.

brucebinary: @MahboobehMohammadi I see. It helps that you find a way to make that intent more clearly communicated. Either better names or just a comment. Future developers reading your code should not need to do detective work on your code.

brucebinary: @MahboobehMohammadi comments should be a last resort.

src/components/CategoricalDisplay.jsx
```diff
@@ -103,7 +103,7 @@ const CategoricalDisplay = ({
     return (
         <div className="cq-categorical-display">
             {renderMobileTitle()}
-            <div className={`cq-lookup-filters ${hideLookupFilter ? 'hide' : ''}`}>
+            <div className={`cq-lookup-filters ${hideLookupFilter ? 'scroll-down' : ''}`}>
```
brucebinary: @MahboobehMohammadi the renaming should include `hideLookupFilter` as well, since we handle the logic to hide it in CSS. Think about someone reading this code in the future. He would be wondering "You want to hide the lookup filter, but you put a class `scroll-down`?".

In short, when ppl read the code in future, if the parts are related, it must read as accordingly.

src/store/CategoricalDisplayStore.js
MahboobehMohammadi: @brucebinary how about `isUserScrolling` ? this pretty much explain it.

brucebinary: That's a better name 👍 . A short comment on the scroll event being triggered when the user keys in something would be helpful as well.

src/components/CategoricalDisplay.jsx
MahboobehMohammadi: @brucebinary Since the class name became `scroll-down` i think ii should use something like `isScrollingDown` or st like that.

brucebinary: sounds good

PR #371, opened 2018-07-09
bug fix: if unlicensed comparison symbol is added, it will load perpe…

brucebinary:
…tually in comparison list

With this PR, when user selects an unlicensed symbol as comparison, it will not be added to chart, and an error notification will be displayed.

### DIFF
--- a/src/feed/Feed.js
+++ b/src/feed/Feed.js
@@ -86,15 +86,20 @@ class Feed {
                     category: 'activesymbol',
                 });
             } else if (errorCode === 'StreamingNotAllowed') {
-                if (!isComparisonChart) {
-                    this._mainStore.chart.setChartAvailability(false);
-                }
                 this._mainStore.notification.notify({
                     text: t.translate('Streaming for [symbol] is not available due to license restrictions', tParams),
                     type: NotificationStore.TYPE_ERROR,
                     category: 'activesymbol',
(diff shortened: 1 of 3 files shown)


src/feed/Feed.js
```diff
@@ -86,15 +86,20 @@ class Feed {
...
                     text: t.translate('Streaming for [symbol] is not available due to license restrictions', tParams),
                     type: NotificationStore.TYPE_ERROR,
                     category: 'activesymbol',
                 });
-                callback({ quotes: [] });
+                let dataCallback = { quotes: [] };
+                if (isComparisonChart) {
+                    // Passing error will prevent the chart from being shown; for
+                    // main chart we still want the chart to be shown, just disabled
+                    dataCallback = { error: errorCode, suppressAlert: true, ...dataCallback };
```
shayanbinary: can it be `call({ error: errorCode, suppressAlert: true, ...dataCallback });` ?  and also remove callback before return

brucebinary: It can, but in both scenarios we also want `callback` to be called. Having `callback` in `if` and `else` block is IMHO more verbose than it needs to be.

No, we can't remove `callback` before return. The callback informs `CIQ.ChartEngine` of the result of fetching the quote data.

PR #374, opened 2018-07-10
Change chart timezone to utc

MahboobehMohammadi:


src/feed/TickHistoryFormatter.js
```diff
@@ -1,16 +1,21 @@
 export class TickHistoryFormatter {
+    static getUTCDate = function (date) {
```
brucebinary: shouldn't this be `static function getUTCDate(date)`?

Also notice the duplicate code below. I think this should be `static function getUTCDate(epoch)`, where you can simply do `TickHistoryFormatter.getUTCDate(tick.epoch)`

src/feed/TickHistoryFormatter.js
```diff
@@ -1,16 +1,21 @@
...
+        const UTCdate = date.toISOString();
+        return UTCdate.substring(0, 10).concat(' ', UTCdate.substring(11, 19));
```
brucebinary: this can be simplified to `UTCdate.substring(0, 19)`. You only need to remove the `Z` at the back, which marks this as UTC time, causing ChartIQ to offset the timezone.

src/feed/TickHistoryFormatter.js
```diff
@@ -1,16 +1,21 @@
...
+        return UTCdate.substring(0, 19);
+    }
+
     static formatHistory(response) {
         const { history, candles } = response;
         if (history) {
             const { times, prices } = history;
             const quotes = prices.map((p, idx) => ({
-                DT: +times[idx] * 1000,
+                Date: this.getUTCDate(+times[idx] * 1000),
```
brucebinary: epoch is in seconds. Do the `* 1000` conversion in `getUTCDate`

src/feed/TickHistoryFormatter.js
MahboobehMohammadi: @brucebinary done!

src/feed/TickHistoryFormatter.js
MahboobehMohammadi: @brucebinary done!

src/feed/TickHistoryFormatter.js
MahboobehMohammadi: @brucebinary yes right.

PR #385, opened 2018-07-16
Light/Dark theme

shayanbinary:
redesign the theme params, add light/dark colors to the theme.
For now, all object of the chart like dialog, menu, list and ... has their own style params


index.html
```diff
@@ -5,7 +5,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1" />
     <meta name="description" content="Analyse the financial markets using beautiful visualisations and unique trading tools, powered by a fast and intuitive interface." />
-    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css" media="screen" />
+    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css?c089caa34345a4b55bd5ba3aa32378de" media="screen" />
```
brucebinary: Pls check the diff before you request for review. Pls. Don't let me repeat this again. Maybe you would want to update the npm script to not overwrite this file, but it takes effort and no one took it up yet.

sass/_themes.scss
```diff
@@ -1,73 +1,470 @@
...
-    controlTextActive:      #e2e2e2,
-    controlBgActive:        rgba(76, 87, 148, 0.2),
-    controlGrid:            #fafafb,
-    controlDialogHead:      mix(#fff, #1c2a35, 3%),
-    controlDialogBody:      mix(#fff, #1c2a35, 7%),
-    tooltipBg:              #1c70cea1,
-    tooltipTextColor:       #fff,
-  ),
+    light: (
+        DefaultMain:            #ffffff,
```
brucebinary: can you cut down the number of variables we need to tweak for the theme? It should not be specific to any component. This is waaaaaay too much. Discuss with Afiq on this. Something along the lines of a dozen colours will do.

src/components/ChartSize.jsx
```diff
@@ -10,12 +10,16 @@ const ChartSize = ({
     zoomIn,
     zoomOut,
 }) => (
-    <div className="ciq-menu cq-chart-size">
-        <div className="cq-menu-btn">
-            <ZoomOutIcon className="ic-icon-with-sub cq-zoom-out" tooltip-title={t.translate('Zoom out')} onClick={zoomOut} />
+    <div className="cq-chart-size-menu">
```
brucebinary: why do you need to create a new CSS class `cq-chart-size-menu`?

src/components/SettingsDialog.jsx
```diff
@@ -108,7 +108,7 @@ const SettingsDialog = ({
                     { isFavoritable &&
                     <StarIcon
                         onClick={onStarClick}
-                        className={`margin ${stared ? 'ciq-active-favorite' : ''}`}
+                        className={`margin icon-favorite ${stared ? 'ciq-active-favorite' : ''}`}
```
brucebinary: our icons are using the convention `ic-{icon name}`. Best to stick to that naming convention; codebase needs to be consistent throughout even though it is written collaboratively.

src/components/CrosshairToggle.jsx
```diff
@@ -9,7 +9,7 @@ const CrosshairToggle = ({
 }) => {
     const CrosshairIcon = [CrosshairOffIcon, CrosshairOnIcon, CrosshairTooltipIcon][state];
     return (
-        <div className="ciq-menu">
+        <div className={`ciq-menu ${(state !== 0) ? 'stxMenuActive' : ''}`}>
```
brucebinary: `stxMenuActive` for toggle? That's very confusing. We styled this without needing `stxMenuActive`. Why is it needed here now?

index.html
shayanbinary: @brucebinary oops I don't really know how I miss this one

sass/_themes.scss
shayanbinary: @brucebinary, of course, we can reduce the variables, but I act this way to separate all modules so that we have less change later.

src/components/ChartSize.jsx
shayanbinary: @brucebinary, as you see in the designs, buttons of chart size, have separate hover effect which conflict with class `ciq-menu`, of course, I can override it, but think it's better to define a container for these two buttons.

src/components/SettingsDialog.jsx
shayanbinary: @brucebinary you are right, I follow the market and name it `ciq-favorite`, is this ok?

src/components/CrosshairToggle.jsx
shayanbinary: @brucebinary  as you see in [the design](https://app.zeplin.io/project/5b3d8a662f4775474e13e26f/screen/5b4f012de88a0ad55ec70460) when crosshair gets active, like other menus it should have the white top border. so why not using the existing classes besides defining new.

sass/_themes.scss
ashkanx: Ideally it shouldn't be like this.
However, we are still in the middle of developing the UI and I'm expecting lots of changes in colours and UI elements. 
So I think separating modules makes sense in this stage until things become stable and we can cut down the variables.

brucebinary: @shayanbinary @ashkanx alright. We keep it as is then.

src/components/CrosshairToggle.jsx
brucebinary: @shayanbinary  better that you do this styling in CSS than in jsx

brucebinary: @shayanbinary actually nvm that.

src/components/ChartSize.jsx
brucebinary: @shayanbinary what separate hover effect? They look the same to me.

shayanbinary: @brucebinary return it back to what it was and handle it by CSS

src/components/CrosshairToggle.jsx
shayanbinary: @brucebinary  how about to handle it by another class name like `crosshair-active`? because its menu should have the top border while crosshair is active.

shayanbinary: @brucebinary As I discuss with Ashkan, crosshair has the wrong design, so I remove what I added here.

sass/styles/chart.scss
```diff
@@ -285,14 +285,75 @@ html, body {
 
 
 /* ------------------------------------------------------- CHART OVERLAY ELEMENTS ------------------------------------------------------ */
+#mSticky{
```
shayanbinary: this is for [the design](https://app.zeplin.io/project/5b3d8a662f4775474e13e26f/screen/5b431ab44db03a8165123a82) of removing Comparison by right click.

PR #390, opened 2018-07-18
Setting dialog style fixes

MahboobehMohammadi:


src/components/SettingsDialog.jsx
```diff
@@ -27,6 +27,7 @@ const SettingsDialog = ({
     onItemChange,
     isFavoritable,
     Dialog,
+    open,
```
brucebinary: why do you need this `open` prop? Seems you only used it to place a  `cq-dialog-active` class. This is not necessary as the parent container has an `open` class when opened.

brucebinary: @MahboobehMohammadi ok I'll just pass this. Seeing that both menu dropdown and dialog the same kind of overlays I think the implementation can be shared, but I'll do this is in a separate branch

PR #393, opened 2018-07-19
Hide doorbell.io network error from console

MahboobehMohammadi:


brucebinary:
As discussed, we will leave the default behaviour as is

PR #395, opened 2018-07-19
Add webpack stylelint plugin

brucebinary:
After much gnashing of teeth, this PR proposes to enforce stylelint so that *.scss files have consistent styling and convention used throughout the code base. Stylelint can be automatically applied to the entire code base via:

    stylelint "**/*.scss" --syntax scss --fix

This PR is also removed a ton of unused css code. It shaves off 27Kb off the css file (162Kb-135Kb).


.travis.yml
```diff
@@ -15,8 +15,8 @@ notifications:
     on_pull_requests: false
     on_success: never
 
-before_install:
-  - yarn install
```
brucebinary: this is unnecessary. Travis will run `yarn install` prior to building by default

PR #400, opened 2018-07-23
Add maxMasterDataSize to prevent it from growing too large

MahboobehMohammadi:


src/store/ChartStore.js
```diff
@@ -179,6 +179,8 @@ class ChartStore {
             yTolerance: 999999, // disable vertical scrolling
         });
 
+        stxx.maxMasterDataSize = 5000;
```
brucebinary: this looks good. I'll be adding it in https://github.com/binary-com/SmartCharts/pull/396 instead of merging this, because this setting needs to sync with the stream manager's caching implementation.

PR #415, opened 2018-07-30
use css animation beside of canvas

shayanbinary:
at last, I use HTML/CSS to make the animation, on the first draw callback I add `span` to the dom and on each draw callback, I just update the `left, top` attributes

### DIFF
--- a/sass/styles/main.scss
+++ b/sass/styles/main.scss
@@ -1497,3 +1497,46 @@ cq-share-dialog {
 .cq-menu-dropdown {
 	line-height: 2.1em;
 }
+
+
+
+
+
+$spot-color: #0298d3;
+
+.cq-marker-spot{
+	width: 6px;
+	height: 6px;
+	border-radius: 50%;
+	cursor: pointer;
+	position: absolute;
+	background: $spot-color;
+	box-shadow: 0 0 0 rgba($spot-color, 0.4);
+	animation: pulse 2s infinite;
+}
+
+@-webkit-keyframes pulse {
+	0% {
+		-webkit-box-shadow: 0 0 0 0 rgba($spot-color, 0.4);
+	}
+	70% {
+		-webkit-box-shadow: 0 0 0 10px rgba($spot-color, 0);
+	}
+	100% {
+		-webkit-box-shadow: 0 0 0 0 rgba($spot-color, 0);
+	}
+}
+@keyframes pulse {
+  0% {
+    -moz-box-shadow: 0 0 0 0 rgba($spot-color, 0.4);
+    box-shadow: 0 0 0 0 rgba($spot-color, 0.4);
+  }
+  70% {
+      -moz-box-shadow: 0 0 0 10px rgba($spot-color, 0);
+      box-shadow: 0 0 0 10px rgba($spot-color, 0);
+  }
+  100% {
+      -moz-box-shadow: 0 0 0 0 rgba($spot-color, 0);
+      box-shadow: 0 0 0 0 rgba($spot-color, 0);
+  }
+}
--- a/src/components/ui/Animation.js
+++ b/src/components/ui/Animation.js
@@ -43,13 +43,10 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
     stx.tickAnimator = easeMachine || new CIQ.EaseMachine(Math.easeOutCubic, 1000);
     let scrollAnimator = new CIQ.EaseMachine(Math.easeInOutCubic, 1000);
 
-    let flashingColors = ['#0298d3', '#19bcfc', '#5dcffc', '#9ee3ff'];
-    let flashingColorIndex = 0;
-    let flashingColorThrottle = 20;
-    let flashingColorThrottleCounter = 0;
-
     let filterSession = false;
     let nextBoundary = null;
+
+    let currentSpotElement;
 
     function initMarketSessionFlags() {
         filterSession = false;
@@ -258,12 +255,6 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
     stx.append('draw', function () {
         if (filterSession) { return; }
         if (this.chart.dataSet && this.chart.dataSet.length && this.mainSeriesRenderer && this.mainSeriesRenderer.supportsAnimation) {
-            if (flashingColorThrottleCounter % flashingColorThrottle === 0) {
-                flashingColorIndex++;
-                flashingColorThrottleCounter = 0;
-            }
-            flashingColorThrottleCounter++;
-
             let context = this.chart.context;
             let panel = this.chart.panel;
             let currentQuote = this.currentQuote();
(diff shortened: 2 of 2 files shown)


src/components/ui/Animation.js
```diff
@@ -275,12 +266,14 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
             if (this.chart.yAxis.left > x &&
                 this.chart.yAxis.top <= y &&
                 this.chart.yAxis.bottom >= y) {
-                if (flashingColorIndex >= flashingColors.length) { flashingColorIndex = 0; }
```
shayanbinary: this code is for previous spot, so no need anymore

src/components/ui/Animation.js
```diff
@@ -275,12 +266,14 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
...
-                context.beginPath();
-                context.moveTo(x, y);
-                context.arc(x, y, 2 + flashingColorIndex * 1.07, 0, Math.PI * 2, false);
-                context.fillStyle = flashingColors[flashingColorIndex];
-                context.fill();
+                if (!currentSpotElement) {
```
shayanbinary: if spot not exists, then add it to DOM, if exists then just update the `Top, Left` attributes.

sass/styles/main.scss
```diff
@@ -1497,3 +1497,46 @@ cq-share-dialog {
...
 	line-height: 2.1em;
 }
+
+
+
+
+
+$spot-color: #0298d3;
+
+.cq-marker-spot{
```
brucebinary: I think just call it `cq-spot` would do. It's not a marker; it follows the current quote

src/components/ui/Animation.js
```diff
@@ -275,12 +266,14 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
...
+                    currentSpotElement = document.createElement('span');
+                    currentSpotElement.className = 'cq-marker-spot';
+                    this.chart.panel.holder.parentElement.appendChild(currentSpotElement);
+                } else {
+                    currentSpotElement.style.top = `${y - 3}px`;
+                    currentSpotElement.style.left = `${x - 3}px`;
```
brucebinary: you need to set `{display: none}` for this when the spot is out of range or when the chart type doesn't support spot.

src/components/ui/Animation.js
```diff
@@ -275,12 +273,9 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
             if (this.chart.yAxis.left > x &&
                 this.chart.yAxis.top <= y &&
                 this.chart.yAxis.bottom >= y) {
-                if (flashingColorIndex >= flashingColors.length) { flashingColorIndex = 0; }
-                context.beginPath();
-                context.moveTo(x, y);
-                context.arc(x, y, 2 + flashingColorIndex * 1.07, 0, Math.PI * 2, false);
-                context.fillStyle = flashingColors[flashingColorIndex];
-                context.fill();
+                currentSpotElement.style.top = `${y - 3}px`;
```
brucebinary: One more thing: can you make `y` an integer? We don't want floating point values in `top`:

`let y = this.pixelFromPrice(price, panel) | 0;`

sass/styles/main.scss
```diff
@@ -1498,3 +1498,46 @@ cq-share-dialog {
...
+	height: 6px;
+	border-radius: 50%;
+	cursor: pointer;
+	position: absolute;
+	background: $spot-color;
+	box-shadow: 0 0 0 rgba($spot-color, 0.6);
+	animation: pulse 1.6s infinite;
+}
+
+@-webkit-keyframes pulse {
```
brucebinary: This is quite a lot of duplicate code btw. Find a way to shorten this down.

This code should be in `chart.scss`. It's also weird to have both `src/app.scss` and `src/main.scss`. I'm thinking we'll need to restructure that soon.

shayanbinary: @binarydevteam I can handle it in this PR or later, what is your idea?

sass/styles/chart.scss
```diff
@@ -975,3 +975,39 @@ html, body {
...
+    height: 6px;
+    border-radius: 50%;
+    cursor: pointer;
+    position: absolute;
+    background: $spot-color;
+    @include box-shadow-raw(0 0 0 rgba($spot-color, 0.6));
+    animation: pulse 1.6s infinite;
+}
+
+@-webkit-keyframes pulse {
```
brucebinary: Can you write a mixin for keyframes (or just find one online)? You have `@-webkit-keyframes` and `@keyframes` having exactly the same code

shayanbinary: oops, you are right, sorry I don't get your point on the first time.

src/components/ui/Animation.js
```diff
@@ -256,31 +253,36 @@ CIQ.Animation = function (stx, animationParameters, easeMachine) {
...
+            currentSpotElement = document.createElement('span');
+            currentSpotElement.className = 'cq-spot';
+            this.chart.panel.holder.parentElement.appendChild(currentSpotElement);
+        }
+
+        currentSpotElement.style.display = 'none';
 
+        if (filterSession) { return; }
+        if (
+            this.layout.timeUnit === 'second'
```
shayanbinary: force animation to just display when then the interval is `1 second`

brucebinary: ... wait. Why?

shayanbinary: @brucebinary think it's not meaningful if animation exist when the interval is something but `1 second`

src/store/CurrentSpotStore.js
```diff
@@ -0,0 +1,51 @@
...
+
+    @observable top = 0;
+    @observable left = 0;
+    @observable show = false;
+
+    get context() { return this.mainStore.chart.context; }
+    get stx() { return this.context.stx; }
+    onContextReady = () => {
+        const self = this;
+        this.stx.append('draw', function () {
```
brucebinary: there is not much code now in smartcharts that uses function context to retrieve its members (`this` in this function would be the chart engine). I'd say from a design standpoint it's easier to read. Consider using:

`this.stx.append('draw', this.updateSpot)`

where `updateSpot` is action bound.

src/store/CurrentSpotStore.js
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            self.updateDisplay(this.layout.periodicity === 1
+                && this.layout.timeUnit === 'second');
+        });
+    }
+
+    @action.bound updateSpot(x, y) {
+        this.top = y;
+        this.left = x > this.left ? x : this.left;
```
brucebinary: this is the source of the issue I screenshot in the card. You can't assume that `x` is always going to be larger because it can be larger when the user scrolls about the chart. Try instead

`Math.abs(x - this.left) > 1`

app/index.jsx
```diff
@@ -76,8 +76,6 @@ const requestForget = streamManager.forget.bind(streamManager);
 
 
 class App extends Component {
-    startingLanguage = 'en';
```
brucebinary: @shayanbinary pls learn to handle merge conflicts better

shayanbinary: screen shot 2018-08-10 at 8 37 23 am
it's not conflict. this is `eslint` error. also by moving this line down, all works fine for me. if there is anything else, I'll be glad to know, so that it would not happen next time.

shayanbinary: that's confusing. because we face 2 different rule for one line. in my side its `startingLanguage should be placed after constructor` and in Travis `startingLanguage should be placed before constructor`

app/index.jsx
```diff
@@ -77,7 +77,6 @@ const requestForget = streamManager.forget.bind(streamManager);
 
 class App extends Component {
     startingLanguage = 'en';
-
```
brucebinary: You don't need to remove this line.

app/index.jsx
shayanbinary: use `yarn install` and it's fixed.

src/store/CurrentSpotStore.js
```diff
@@ -0,0 +1,53 @@
...
+            if (!currentQuote) { return; }
+            const price = currentQuote.Close;
+            let x = this.stx.pixelFromTick(currentQuote.tick, chart);
+            if (chart.lastTickOffset) { x += chart.lastTickOffset; }
+            const y = this.stx.pixelFromPrice(price, panel);
+            if (chart.yAxis.left > x
+                && chart.yAxis.top <= y
+                && chart.yAxis.bottom >= y) {
+                this.top = y;
+                this.left = (Math.abs(x - this.left) > 1) ? x : this.left;
```
brucebinary: @shayanbinary if you are using doubles, then you can just set `this.left = x` right?

PR #426, opened 2018-08-01
provide library user option to disable market selector

brucebinary:


sass/components/_chart-title.scss
```diff
@@ -1,122 +1,123 @@
 .cq-chart-title {
+    display: inline-block;
     top: 1em;
     z-index: 41;
     margin: 0.2em 0.8em;
 
-    &.stxMenuActive {
```
brucebinary: most of the styling here is only used in chart controls, so I moved them over there instead. This prevents chart control styles from bleeding into the chart title

sass/components/_chart-title.scss
```diff
@@ -1,122 +1,123 @@
...
-                height: $large-icon-size;
-                width: $large-icon-size;
-                margin-right: 12px;
-                transform: scale(1.3);
-
-                > svg {
-                    height: $large-icon-size;
-                }
+            > .ic-icon.cq-symbol-dropdown {
+                transform: rotate(180deg);
```
brucebinary: In general I'm opposed to using `!important` as much as possible in the code base. Either make the query more specific or restructure the classes to give it precedence.

src/components/ChartTitle.jsx
```diff
@@ -13,44 +13,54 @@ const ChartTitle = ({
     isPriceUp,
     currentSymbol,
     Menu,
-    CategoricalDisplay,
+    MarketSelector,
```
brucebinary: this is to remove eslint's shadow warnings. IMHO it also expresses intent more clearly

src/components/ChartTitle.jsx
```diff
@@ -13,44 +13,54 @@ const ChartTitle = ({
...
     AnimatedPrice,
     onCloseMenu,
-    isMobile,
```
brucebinary: `isMobile` is retrieved from the chart store in `MenuStore`, so this prop is redundant.

brucebinary:
@shayanbinary which part did you not understand? Some of the big change here is moving the chart control menu buttons style into `chart-control.scss`, because I noticed that it was affecting the chart title, but patched up with `!important`. This makes for confusing code. If it is separate, then style them separately.

PR #428, opened 2018-08-02
Browser support

shayanbinary:


app/SupportDetection.jsx
```diff
@@ -0,0 +1,108 @@
...
+
+    componentDidCatch(error, info) {
+        this.setState({ hasError: true });
+        console.error(error, info);
+    }
+
+    render() {
+        const NotGuaranteed = (
+            <div className="cq-not-guaranteed">
+                <h1>We are  NOT guaranteed</h1>
```
zakibinary: This is a placeholder I assume?

app/SupportDetection.jsx
```diff
@@ -0,0 +1,108 @@
...
+                <p>Binary.com does not guaranteed your browser. Please update your browser.</p>
```
zakibinary: Same here. Maybe we can get our copy writer to have a look for a better message.

app/SupportDetection.jsx
shayanbinary: @zakibinary by @ashkanx suggestion, on the case we can not recognize browser, we should a message to the user, which `We can NOT guarantee your browser`

zakibinary: Yep, I agree that the message should be shown, but perhaps we have our copy writer come up with a more suitable message to show to users in case the browser is not supported/recognized.

package.json
```diff
@@ -106,13 +110,15 @@
...
+    "detect-browser": "^3.0.0",
     "event-emitter-es6": "^1.1.5",
     "lodash.debounce": "^4.0.8",
     "prop-types": "^15.6.0",
     "react-perfect-scrollbar": "^1.1.1",
     "react-transition-group": "^2.3.1",
     "resize-observer-polyfill": "^1.5.0",
-    "robust-websocket": "^0.3.0"
+    "robust-websocket": "^0.3.0",
+    "stylelint-no-unsupported-browser-features": "^3.0.1"
```
shayanbinary: as we use Autoprefixer, maybe this plugin is useless

package.json
```diff
@@ -87,6 +88,9 @@
     "perfect-scrollbar": "^1.3.0",
     "po-loader": "^0.4.1",
     "po2json": "^0.4.5",
+    "postcss-import": "^12.0.0",
+    "postcss-loader": "^2.1.6",
+    "postcss-preset-env": "^5.3.0",
```
shayanbinary: these are for Autoprefixer library

browserslist
```diff
@@ -0,0 +1,3 @@
+safari > 10
+ie > 9
```
brucebinary: we are not supporting ie.

package.json
brucebinary: if you use autoprefixer, you shouldn't need most of the `@mixin` that just prefix `mox-` or `webkit-`

app/SupportDetection.jsx
brucebinary: Do we need to translate that message? THere's no way of setting language tho

ashkanx: use this:
<h1>Incompatible browser</h1>
<p>Binary.com does not recognise or support your current browser. Please *update your browser* immediately.</p>

shayanbinary: @ashkanx Any special design? for example for the button `update your browser`.

package.json
shayanbinary: @brucebinary I should remove `@mixin` or just remove `webkit-` props on the mixin? If I am going to remove `mixin` then I have to modify a lot of CSS file.

app/SupportDetection.jsx
zakibinary: Hi @shayanbinary , I sent you a link for the design we're using on smart-trader.

package.json
brucebinary: @shayanbinary sure go for it. That's why we want to introduce autoprefixer in the first place.

index.html
```diff
@@ -5,7 +5,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="user-scalable=no, width=device-width, initial-scale=1, maximum-scale=1" />
     <meta name="description" content="Analyse the financial markets using beautiful visualisations and unique trading tools, powered by a fast and intuitive interface." />
-    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css" media="screen" />
+    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css?af970fc251e117659a02ccec6039bdd7" media="screen" />
```
brucebinary: this should not be in the commit.

sass/_variables.scss
```diff
@@ -60,39 +55,10 @@ $small-icon-size: 16px;
...
 }
 
-@mixin transition($transition...) {
-    -webkit-transition: $transition;
-    -moz-transition: $transition;
-    -ms-transition: $transition;
-    transition: $transition;
-}
-
 @mixin border-radius($radius : $soft-corners) {
```
brucebinary: if autoprefix prepends automatically you don't need this mixin

index.html
shayanbinary: That's a shame for me, sorry.

webpack.config.js
```diff
@@ -30,6 +30,16 @@ const config = {
                         loader: 'css-loader',
                         options: { sourceMap: true },
                     }, {
+                      loader: 'postcss-loader',
+                      options: {
+                        ident: 'postcss',
```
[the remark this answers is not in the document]
brucebinary: you are so sure? I don't see that documented in their documentation. Even in binary-static's source code, the autoprefixer is specifically defined in `postcss.js`

app/SupportDetection.jsx
```diff
@@ -0,0 +1,113 @@
...
+            } else {
+                this.setState({ unknown: true });
+            }
+        }
+    }
+
+    componentDidMount() {
+        const { onMessage } = this.props;
+        if (this.state.unknown && onMessage) {
+            onMessage({
```
brucebinary: @shayanbinary why do you need SupportDetection to intercept the chart messages?...

brucebinary: I also resolved the merge conflicts for ya. Those were quite complicated.

PR #430, opened 2018-08-02
make notification container relative in mobile

shayanbinary:


app/_ciq-notification.scss
```diff
@@ -82,3 +82,10 @@
     }
 }
 
+.smartcharts-mobile {
+    .cq-notifications {
+        position: relative;
+        right: 0px;
+        top: 0px;
```
MahboobehMohammadi: @shayanbinary can we have something like a top:-25px so we remove the space between notification and the market selector? just a suggestion though.

shayanbinary: @MahboobehMohammadi the problem is that `notification` div is under the `asset information` div if I move it up, it go on the asset information

MahboobehMohammadi: @shayanbinary fair point.its okay though

MahboobehMohammadi:
LGTM

app/_ciq-notification.scss
shayanbinary: @MahboobehMohammadi yeah you are right, the margin belongs to `comparison` container. which I remove it on mobile mode.

MahboobehMohammadi: @shayanbinary yep. much better,thanks.

MahboobehMohammadi:
LGTM.

PR #433, opened 2018-08-03
Prevent user from moving the indicator panel above the main chart

MahboobehMohammadi:
…chart


src/store/ChartStore.js
```diff
@@ -224,6 +224,20 @@ class ChartStore {
...
         stxx.append('deleteHighlighted', this.updateComparisons);
+
+        stxx.append('adjustPanelPositions', function () {
+            let index = 0;
+            for (const x in stxx.panels) {
+                // Hide arrow up for first study
+                if (index == 1) {
+                    const panel = stxx.panels[x];
+                    panel.up.style.display = 'none';
+                    break;
```
shayanbinary: @MahboobehMohammadi I think this `break` is useless

src/store/ChartStore.js
```diff
@@ -224,6 +224,20 @@ class ChartStore {
         this.holderStyle = stxx.chart.panel.holder.style;
 
         stxx.append('deleteHighlighted', this.updateComparisons);
+
+        stxx.append('adjustPanelPositions', function () {
+            let index = 0;
+            for (const x in stxx.panels) {
```
shayanbinary: @MahboobehMohammadi also it's work fine. but think it's better to prevent loop in each `adjustPanelPositions`, Could you please handle it in another way?

MahboobehMohammadi: @shayanbinary this loop only happen twice, for index 0 and 1, and  to find the panel inside the stx.pannels, i need panel name and dont have it here, its why im using this for loop, actually its the way chartiq.js itself access panels.
http://documentation.chartiq.com/CIQ.ChartEngine.Panel.html#Panel

shayanbinary: @MahboobehMohammadi all look good to me. I am just thinking about something like `const Index = Object.keys(stxx.panels)[1];const indicate = sttx.panels[index];indicate.up.style.display = 'none';`. just to use less code.

src/store/ChartStore.js
shayanbinary: @MahboobehMohammadi you are right. but I think it's not so good to make a loop then break it because u just need index 0 and 1.

src/store/ChartStore.js
MahboobehMohammadi: @shayanbinary you got it

src/store/ChartStore.js
```diff
@@ -224,6 +224,12 @@ class ChartStore {
         this.holderStyle = stxx.chart.panel.holder.style;
 
         stxx.append('deleteHighlighted', this.updateComparisons);
+
+        stxx.append('adjustPanelPositions', function () {
+            const panel = Object.keys(stxx.panels)[1];
+            stxx.panels[panel].up.style.display = 'none';
```
shayanbinary: @MahboobehMohammadi LGTM. just one simple modification. check to see if `stxx.panels[panel]`is defined or not. maybe the user does not select any indicator.

src/store/ChartStore.js
```diff
@@ -223,6 +223,14 @@ class ChartStore {
         this.holderStyle = stxx.chart.panel.holder.style;
 
         stxx.append('deleteHighlighted', this.updateComparisons);
+
+        stxx.append('adjustPanelPositions', function () {
+            const panel = Object.keys(stxx.panels)[1];
```
brucebinary: A comment would be helpful for future developers

PR #434, opened 2018-08-03
Update to Mobx 5

brucebinary:


package.json
```diff
@@ -116,7 +116,7 @@
   },
   "peerDependencies": {
     "babel-polyfill": "^6.26.0",
-    "mobx": "^4.2.0",
+    "mobx": "^5.0.3",
     "mobx-react": "^5.1.2",
```
morteza-binary: @brucebinary Would you plz update `mobx-react` to `5.2.3` since we have updated it in binary-static.

brucebinary: @morteza-binary sure

PR #441, opened 2018-08-07
Add ability to pass color code for barriers

MahboobehMohammadi:


src/components/Barrier.jsx
```diff
@@ -17,9 +17,13 @@ const Barrier = ({
     hidePriceLines,
     lineStyle,
     isInitialized,
-}) => (isInitialized &&
+}) => {
+    const chartContainer = document.getElementsByClassName('chartContainer')[0];
+    chartContainer.style.setProperty('--shade-color', shadeColor);
```
brucebinary: this sets the shade color as part of the chartContainer. This property should be contained within the barrier itself. Preferably not using `getElementsByClassName` inside the render method as you are querying the DOM tree each time, and preferably setting the style property via react than manipulating the DOM directly.

MahboobehMohammadi: @brucebinary done!

app/test.jsx
```diff
@@ -134,6 +137,16 @@ class App extends React.Component {
...
+                        Choose barrier bg color:&nbsp;
+                        <select defaultValue="" id="barrierBGColor" onChange={this.onColorChange}>
+                            <option value="GREEN">GREEN</option>
+                            <option value="RED">RED</option>
+                            <option value="YELLOW">YELLOW</option>
+                            <option value="ORANGERED">ORANGERED</option>
+                            <option value="PURPLE">PURPLE</option>
+                            <option value="BLUE">BLUE</option>
+                            <option value="DEEPPINK">DEEPPINK</option>
+                        </select>
```
MahboobehMohammadi: @brucebinary I just added some color here to test barrier color changed in realtime.

src/store/BarrierStore.js
```diff
@@ -4,7 +4,6 @@ import ShadeStore from './ShadeStore';
 import PendingPromise from '../utils/PendingPromise';
 
 export default class BarrierStore {
-    static get SHADE_COLOR_RED() { return 'red'; }
     static get SHADE_COLOR_GREEN() { return 'green'; }
```
brucebinary: Just remove this as well. In `Barrier.jsx` you can use default params to set `green` (like how `color` prop is set

app/test.jsx
brucebinary: Works pretty nicely. Nice job (:

src/store/BarrierStore.js
```diff
@@ -15,7 +13,8 @@ export default class BarrierStore {
 
     static get BARRIER_CHANGED() { return 'BARRIER_CHANGED'; }
 
-    @observable shadeColor = BarrierStore.SHADE_COLOR_GREEN;
+    @observable shadeColor;
+    @observable color;
```
MahboobehMohammadi: @brucebinary barrier line color can change in realtime too.

PR #459, opened 2018-08-14
Add id to smartchart

MahboobehMohammadi:


src/store/ChartState.js
```diff
@@ -38,14 +39,15 @@ class ChartState {
     }
 
     saveLayout() {
+        if (!this.chartId) return;
         const layoutData = this.stxx.exportLayout(true);
         const json = JSON.stringify(layoutData);
-        CIQ.localStorageSetItem(`layout-${this.id}`, json);
+        CIQ.localStorageSetItem(`${this.chartId}-layout-${this.id}`, json);
```
brucebinary: with this proposed change, we don't use `this.id` anymore

src/components/Chart.jsx
```diff
@@ -50,6 +50,7 @@ class Chart extends Component {
 
     render() {
         const {
+            chartId, // eslint-disable-line no-unused-vars
```
brucebinary: why did you disable this?

src/store/ViewStore.js
```diff
@@ -24,12 +24,13 @@ export default class ViewStore {
     get loader() { return this.mainStore.loader; }
 
     onContextReady = () => {
-        const views = createObjectFromLocalStorage('cq-views');
+        const views = createObjectFromLocalStorage(`${this.chartId}-cq-views`);
```
brucebinary: `cq-view` should be shared among all charts, I think. Currently how it is done should stay that way

src/store/ChartState.js
```diff
@@ -105,33 +107,35 @@ class ChartState {
...
+        const drawings = createObjectFromLocalStorage(`${this.chartId}-${this.stxx.chart.symbol}`);
         if (drawings) {
             this.stxx.importDrawings(drawings);
             this.stxx.draw();
         }
     }
 
     restorePreferences() {
-        const pref = createObjectFromLocalStorage(`preferences-${this.id}`);
+        const pref = createObjectFromLocalStorage(`${this.chartId}-preferences-${this.id}`);
```
MahboobehMohammadi: @brucebinary what about here. i didnt know what its used for so i didnt remove it.

src/components/Chart.jsx
MahboobehMohammadi: @brucebinary  its not being used directly in chart.jsx so eslit creates a error.
`  53:13  error  'chartId' is assigned a value but never used  no-unused-vars`

src/store/ChartState.js
brucebinary: me neither. Have a look at the docs and see what it does? If it doesn't save anything then just remove it. This is from the demo code

src/components/Chart.jsx
brucebinary: if it not being used why do you include it?

src/store/ChartState.js
```diff
@@ -27,7 +27,8 @@ class ChartState {
         this.stxx.addEventListener('preferences', this.savePreferences.bind(this));
     };
 
-    @action.bound updateProps({ settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch }) {
+    @action.bound updateProps({ chartId, settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch }) {
+        this.chartId = chartId;
```
MahboobehMohammadi: @brucebinary Here im using it.

src/components/Chart.jsx
MahboobehMohammadi: @brucebinary Its being used in updateProps in chartstate.js.

src/store/ChartState.js
MahboobehMohammadi: @brucebinary  As far as i understood we have one layout for each chart so im going to remove `this.id` from here. dont think its needed since now we have an id for each chart.

src/store/ChartState.js
```diff
@@ -105,33 +107,35 @@ class ChartState {
...
+            CIQ.localStorage.removeItem(`${this.chartId}-${symbol}`);
         } else {
-            CIQ.localStorageSetItem(symbol, JSON.stringify(obj));
+            CIQ.localStorageSetItem(`${this.chartId}-${symbol}`, JSON.stringify(obj));
         }
     }
 
     restoreDrawings() {
-        const drawings = createObjectFromLocalStorage(this.stxx.chart.symbol);
+        const drawings = createObjectFromLocalStorage(`${this.chartId}-${this.stxx.chart.symbol}`);
```
MahboobehMohammadi: @brucebinary drawings need separate keys, since they are specific to each symbol.

src/store/ViewStore.js
MahboobehMohammadi: @brucebinary  views will be shared between all instances.

src/components/Chart.jsx
MahboobehMohammadi: @brucebinary yea. i realy dont need it!

src/store/ChartState.js
brucebinary: @MahboobehMohammadi ...so we can make them specific to each symbol specific to each chart.

MahboobehMohammadi: @brucebinary  yes. its specefic for each symbol specific to each chart. its why we use both `chartId` and `stxx.chart.symbol` in the key.

src/store/ChartState.js
MahboobehMohammadi: @brucebinary  Im gonna remove it.Yes this never appears in local storage. i dont think it has any use.

src/store/ChartState.js
brucebinary: Ok as discussed, we will use separate keys

src/store/SingletonViews.js
```diff
@@ -0,0 +1,23 @@
...
+        }
+        this.mainStore = mainStore;
+        this.views = createObjectFromLocalStorage('cq-views') || [];
+        return singletonViews;
+    }
+
+    @observable views = [];
+
+    updateLocalStorage = () => {
+        if (!this.mainStore.state.chartId) return;
```
brucebinary: we should still let users save templates even though chart ID is not defined. Otherwise saving template in statements page won't do anything.

Also don't pass mainStore to a singleton. mainStore instance is unique to each smartchart instance, whereas singletons have only one copy for the entire life time of the app.

app/index.jsx
```diff
@@ -74,14 +76,13 @@ const requestAPI = connectionManager.send.bind(connectionManager);
...
 
-
 class App extends Component {
     startingLanguage = 'en';
 
     constructor(props) {
         super(props);
         this.notifier = new ChartNotifier();
-        const settings = createObjectFromLocalStorage('smartchart-setting');
+        const settings = createObjectFromLocalStorage(`${chartId}-smartchart-setting`);
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary  hmm. i wanted to use same name conventions for all ids by okay.

src/store/ChartState.js
```diff
@@ -38,14 +38,15 @@ class ChartState {
...
         const layoutData = this.stxx.exportLayout(true);
         const json = JSON.stringify(layoutData);
-        CIQ.localStorageSetItem(`layout-${this.id}`, json);
+        CIQ.localStorageSetItem(`${this.chartId}-layout`, json);
     }
 
     // returns false if restoring layout fails
     restoreLayout() {
-        let layoutData = createObjectFromLocalStorage(`layout-${this.id}`);
+        let layoutData = createObjectFromLocalStorage(`${this.chartId}-layout`);
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary We'll still be having, the chartId prefix for drawings, and i think its better for reading.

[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary okay, so i set the chartId to 1 in app and  the layout to `Layout-1` for drawings the name will be `1-symbol`  or `symbol-1` ?

brucebinary:
@MahboobehMohammadi fyi merge conflict is due to fix here: https://github.com/binary-com/SmartCharts/pull/461/commits/de933141a511105535f69c47066ede06f9e940d4

src/store/ViewStore.js
```diff
@@ -11,6 +11,7 @@ export default class ViewStore {
         when(() => this.context, this.onContextReady);
     }
 
+    staticViews = observable.box(ViewStore).get();
```
brucebinary: very interesting design choice... could you explain your thought process on this?

src/store/ViewStore.js
```diff
@@ -10,9 +10,10 @@ export default class ViewStore {
         this.ViewsMenu = this.menu.connect(Menu);
         when(() => this.context, this.onContextReady);
     }
+
+    staticViews = observable.box(ViewStore).get();
```
brucebinary: very interesting design choice... could you explain your thought process on this?

brucebinary:
@MahboobehMohammadi I'm not quite sure why you can't get static observables to work, but here's a PR that uses this: https://github.com/MahboobehMohammadi/SmartCharts/pull/1

PR #464, opened 2018-08-18
set to update chart when symbol status changed

shayanbinary:


src/store/ChartStore.js
```diff
@@ -219,6 +218,8 @@ class ChartStore {
                     this.changeSymbol(this.state.symbol, this.state.granularity);
                 }
             });
+
+            this.onMarketClosedStatus();
```
shayanbinary: bring this down so that `onMarketClosedStatus` run after the code apply the active symbols, so the checking of closed status do not run again on first initial

brucebinary:
@shayanbinary this PR is tied to the old code base and no longer relevant.

PR #470, opened 2018-08-20
Categorical correct scroll

shayanbinary:
Detect active menu, by measuring the distance of menus from top of scroll panel element


src/store/CategoricalDisplayStore.js
```diff
@@ -82,32 +82,19 @@ export default class CategoricalDisplayStore {
         if (this.pauseScrollSpy || !this.scrollPanel) { return; }
         if (this.filteredItems.length === 0) { return; }
 
+        const ActiveViewHeight = this.scrollPanel.clientHeight * 1 / 5;
```
shayanbinary: If the menu comes to 1/5 of view height, then it should be active

PR #478, opened 2018-08-22
browser detection component

shayanbinary:


app/browser-detection.js
```diff
@@ -0,0 +1,226 @@
...
+        cqA.href = 'http://outdatedbrowser.com/';
+        cqA.innerHTML = 'Update Browser';
+        cqNotSupported.appendChild(cqA);
+        
+        if (cqRoot.nextSibling) {
+            cqRoot.parentNode.insertBefore(cqNotSupported, cqRoot.nextSibling);
+        } else {
+            cqRoot.parentNode.appendChild(cqNotSupported);
+        }
+        cqRoot.parentNode.removeChild(cqRoot);
```
shayanbinary: sometimes we show `not supported` page because of CSS or any critical issue, so as I didn't find any safe way to warranty stop running rest of codes, I decide to stop it by removing `<div id="root" />`.

app/browser-detection.js
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    window.onerror = function(message, source, lineno, colno, error){
+        if (
+                (
+                    message.indexOf('[mobx] MobX 5+ requires Proxy objects') !== -1 ||
+                    message.indexOf('not supported browser') !== -1
```
shayanbinary: check error messages, perhaps a simple error happens that is not related to our browser support check

app/index.jsx
```diff
@@ -124,21 +125,25 @@ class App extends Component {
         const { settings, isConnectionOpened, symbol } = this.state;
 
         return (
-            <SmartChart
-                id={chartId}
-                symbol={symbol}
+            <SupportDetection
```
shayanbinary: as decouple detection code perhaps I should remove this. agree?

app/browser-detection.js
```diff
@@ -0,0 +1,226 @@
...
+        var os = browser.os;
+
+        if (
+            SupportedBrowsers[os] && 
+            SupportedBrowsers[os][name] !== undefined) {
+            if(version < SupportedBrowsers[os][name]){
+                throw new Error('not supported browser');
+            }
+        } else {
+            alert('Please update your browser to the latest version.');
```
shayanbinary: because I don't have access to `notification` component so the only way is to `alert` the message

brucebinary: @shayanbinary why are you using `alert`? We have a design from the UX team to display some page is the browser is not supported. Use that instead

app/index.jsx
brucebinary: yep

app/browser-detection.js
```diff
@@ -0,0 +1,256 @@
+function browserDetect(rootEleId, SupportedBrowsers){
```
brucebinary: what's this script here. Did you wrote this?

webpack.config.js
```diff
@@ -93,6 +93,10 @@ const config = {
         }),
         new MiniCssExtractPlugin({ filename: 'smartcharts.css' }),
         new StyleLintPlugin(),
+        new CopyWebpackPlugin([
```
shayanbinary: @brucebinary Can I leave this piece of code here or I should put in under `if (production) {` ?

ashkanx:
@shayanbinary conflicts

shayanbinary:
@ashkanx Done.

PR #480, opened 2018-08-23
Fix download problem in ios

MahboobehMohammadi:


src/store/ShareStore.js
```diff
@@ -38,22 +43,25 @@ export default class ShareStore {
 
     @action.bound downloadPNG() {
         this.isLoadingPNG = true;
+        const newTab = this.createNewTab();
```
MahboobehMohammadi: I must create the new tab in button click otherwise ios will block it.

src/utils/index.js
```diff
@@ -78,21 +78,33 @@ export function sameBar(bar1, bar2) {
...
         document.body.removeChild(link);
+        return;
+    }
+    if (newTab) {
+        if (type === 'image/png;') {
+            newTab.document.write(`<iframe src="${url}" frameborder="0" style="border:0; top:0px; left:0px; bottom:0px; right:0px; width:100%; height:100%;" allowfullscreen></iframe>`);
+        } else {
+            const lines = content.split('\n');
+            for (let i = 0; i < lines.length; i++) {
+                newTab.document.write(`${lines[i]}<br/>`);
```
MahboobehMohammadi: Printing file lines into browser so it work on both safari and chrome browsers in ios phones

PR #481, opened 2018-08-24
Delayed Feed

brucebinary:
# This PR adds the following features:

 - **Real time market open close updates**. When market open/closes, the chart title and comparison panels will update immediately. If there is a current market streaming, it will notify if the market closes. If an active market goes from close to open, the chart will refresh to stream the market and notify the user accordingly.
 - **SmartCharts is now aware in advance of market status**. Prior to this implementation, smartcharts makes a `tick_history` request to query whether market is closed, restricted, or has realtime quotes. Now smartcharts knows this in advance and will not need to make any calls to backend for this information.
 - **Initial support for delayed feeds**. Delayed feeds (singapore index, bombay index, etc...) will now show a notification indicating it is a delayed feed, and will update with the same interval as the delay. In the future we will be able to stream delayed feeds, or have tick data come in every few seconds, depending on whether binary's API can accommodate these requests.

# This PR introduces a few refactors (for developers only):

## SmartCharts BinaryAPI Interface Changes

All classes that communicate with binary API is placed in `binaryapi`: `BinaryAPI`, `TradingTimes`, `ActiveSymbols`.

#### TradingTimes

`TradingTimes` interfaces with the `trading_times` API, informing smartcharts whether a feed is unavailable (`isFeedUnavailable`), delayed (`getDelayedMinutes`), or closed (`isMarketOpened`). Is also has an `onMarketOpenCloseChanged` event that fires each time any market open or closes, passing a `changes` object which contains all the markets that has changed.

#### ActiveSymbols

`ActiveSymbols` interfaces with `active_symbols`, but has access to an instance of `TradingTimes` to update the symbols `exchange_is_open` field when the market open/closes. The updates happen in a `@computed` method, which causes the chart title and comparisons panel to be updated with close labels the instant the market open/closes.

## Feed

Subscription logic has now been refactored to `Subscription` class, which Feed uses as an abstraction layer for `RealtimeSubscription` and `DelayedSubscription`.

## Notifications

All chart notifications fired from SmartCharts now goes through `Notifier`, so that all notification messages can edited from one place.


src/binaryapi/TradingTimes.js
```diff
@@ -0,0 +1,163 @@
...
+
+            const periodicUpdate = async () => {
+                const changed = this._updateMarketOpenClosed();
+                if (Object.keys(changed).length > 0) {
+                    this._emitter.emit(TradingTimes.EVENT_MARKET_OPEN_CLOSE_CHANGE, changed);
+                }
+                let nextUpdate = this._nextUpdateDate();
+
+                if (!nextUpdate) {
+                    const now = new Date();
```
MahboobehMohammadi: @brucebinary you could move this to `_nextUpdateDate`,  it's easier to read, if the `nextdate` was set inside th `_nextUpdateDate`

brucebinary: @MahboobehMohammadi good point... 

Putting it inside `_nextUpdateDate` causes the unit tests to fail though. Not quite sure what went wrong there. Some race condition thing happening. I'll take a look again tmrw.

PR #487, opened 2018-08-28
Make the dialog width responsive to aviod truncating it when browser …

MahboobehMohammadi:
…width changes


sass/components/_categorical-display.scss
```diff
@@ -348,6 +348,19 @@
     }
 }
 
+.smartcharts-desktop {
+    &.smartcharts-480 {
```
brucebinary: you could fuse the bottom code block with `&.smartcharts-480, &.smartcharts-900`

I'm still passing to QA to see if it solves that they asked for

MahboobehMohammadi: @brucebinary but their width is different, its `450px` for `.smartcharts-480` and `550px` for `.smartcharts-900`

PR #490, opened 2018-08-30
modify font of smartchart

shayanbinary:


sass/app.scss
```diff
@@ -19,12 +19,11 @@ for building additional themes.
 
  */
 /* TODO: deal with translations on all :after and :before content */
-@import url('https://fonts.googleapis.com/css?family=Roboto:400,700,500,100,300,400italic|Roboto+Condensed:400,700,300');
```
brucebinary: did you look at the `constants.scss` file from binary-static that I link in the card? Take use their fonts and their color palette, not copy paste the contents of the this URL.

PR #491, opened 2018-08-30
Slider Refactor

shayanbinary:


brucebinary:
@shayanbinary could you have a look something like https://github.com/jonathantneal/postcss-input-range and see if will cater to our needs? I don't want any vendor specific styling if possible.

shayanbinary:
@brucebinary  Just test it, but still has the problem on Edge.

PR #494, opened 2018-09-03
Prevent chart from moving up when aggregate type selected

MahboobehMohammadi:


src/store/ChartTypeStore.js
```diff
@@ -173,7 +174,9 @@ export default class ChartTypeStore {
             this.stx.chart.tension = 0;
             delete this.stx.layout.tension;
             if (this.aggregates[type.id]) {
+                this.stx.chart.baseline.userLevel = false;
                 this.stx.setAggregationType(type.id);
+                this.stx.chart.baseline.userLevel = true;
```
MahboobehMohammadi: i dont know when `basseline` need to be adjust so i just set it back to its default value after chart type changes.

src/store/ChartTypeStore.js
```diff
@@ -173,7 +173,9 @@ export default class ChartTypeStore {
             this.stx.chart.tension = 0;
             delete this.stx.layout.tension;
             if (this.aggregates[type.id]) {
+                this.stx.chart.baseline.userLevel = false;
```
brucebinary: could you add a comment to explain why these 2 lines of code are needed?

PR #509, opened 2018-09-06
Color palette fonts

shayanbinary:


sass/_variables.scss
```diff
@@ -27,11 +26,10 @@ $small-icon-size: 16px;
     transition: pointer-events 0.2s cubic-bezier(0.64, 0.04, 0.35, 1), transform 0.25s cubic-bezier(0.64, 0.04, 0.35, 1), opacity 0.17s cubic-bezier(0.64, 0.04, 0.35, 1), z-index 0.17s cubic-bezier(0.64, 0.04, 0.35, 1);
 }
 
-@mixin svg-icon-color($color) {
+@mixin svg-icon-color($color, $important:false) {
```
brucebinary: do you really need this `$important`? You only use this once. I think you can do without it.

sass/icons/chart settings/chart types/spline/ic-spline-normal.svg
```diff
@@ -1,10 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <style>
+        .icon-spline{
+            stroke: none !important;
```
brucebinary: Shouldn't this be in scss instead?

sass/icons/chart settings/chart types/baseline/ic-baseline-normal.svg
```diff
@@ -1,12 +1,12 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
     <g fill="none" fill-rule="evenodd">
```
brucebinary: I find it odd you need to use a class `no-stroke`. In `<g />` you can set `stroke="none"  fill="none"` right?

brucebinary:
Ok. As discussed I'll merge this first since it's required for another task. Do look into my comment on removing the need for adding style classes to those icons

PR #521, opened 2018-09-13
Fix png download problem in ios

MahboobehMohammadi:


sass/styles/chart.scss
```diff
@@ -978,4 +978,8 @@ $spot-color: #0298d3;
             margin-left: -2px !important;
         }
     }
+    .ciq-chart.ciq-screenshot .stx-btn-panel {
+        //CSS images cause cross origin security error when export png
+        display: none !important;
```
brucebinary: nitpick: you can put this under `.ciq-chart.ciq-screenshot` in line 913

MahboobehMohammadi: @brucebinary I want this only for mobiles

brucebinary: @MahboobehMohammadi ah I see. Doesn't hurt to put it at both places I think. I won't miss it not being in the desktop screenshot

MahboobehMohammadi: @brucebinary yup. it will. because it hides the pngs for a sec and it doesnt looks good in desktop.

PR #526, opened 2018-09-18
Use server time instead of local time

MahboobehMohammadi:


src/utils/ServerTime.js
```diff
@@ -0,0 +1,60 @@
...
+    constructor(api) {
+        this._api = api;
+        this.init();
+    }
+
+    init() {
+        if (!this.clockStarted) {
+            clearInterval(this.getTimeInterval);
+            this.requestTime();
+            this.getTimeInterval = setInterval(this.requestTime.bind(this), 60000);
```
brucebinary: why are you updating every 60 seconds? Binary-static updates every 30 seconds.

src/binaryapi/TradingTimes.js
```diff
@@ -28,7 +28,7 @@ class TradingTimes {
                 let nextUpdate = this._nextUpdateDate();
 
                 if (!nextUpdate) {
-                    const now = new Date();
+                    const now = this._serverTime.getLocalDate();
```
brucebinary: Try and see if you can change it to only use `getUTCDate`. With this you should only need `get`.

src/binaryapi/TradingTimes.js
```diff
@@ -28,7 +28,7 @@ class TradingTimes {
...
                     const getUpdateDate = () => new Date(`${this.lastUpdateDate}T00:00:00Z`);
```
brucebinary: I suspect there is a bug in this line of code. I've set it to use UTC time but then calculate the difference using local time. Changing everything to use UTC time via `ServerTime` should resolve this issue.

src/store/TimeperiodStore.js
```diff
@@ -65,7 +65,7 @@ export default class TimeperiodStore {
 
             const dataSet = stx.chart.dataSet;
             if (dataSet && dataSet.length !== 0) {
-                const now = new Date();
+                const now = this.mainStore.chart.serverTime.getLocalDate();
```
brucebinary: this should be changed to use UTC. Then the timezone offset below should no longer be required.

src/utils/ServerTime.js
```diff
@@ -0,0 +1,60 @@
...
+
+class ServerTime {
+    clockStarted = false;
+
+    constructor(api) {
+        this._api = api;
+        this.init();
+    }
+
+    init() {
```
brucebinary: we can make `ServerTime` a singleton by passing `api` into `init` instead. This way we don't have to pass a `ServerTime` instance around. It's interesting to note that in their original implementation, binary-static returns undefined if the server time has not been retrieved.

I think we can get around this by making `init` async, and throwing an Error if we try to retrieve the time before the server has responded.

src/binaryapi/TradingTimes.js
MahboobehMohammadi: @brucebinary yup. youre right. it must be in utc

src/store/TimeperiodStore.js
MahboobehMohammadi: @brucebinary  `serverTime.getUTCDate` returns an string and `dataSet[dataSet.length - 1].DT` is a date object so realy cant minus them, if thats what you meant

brucebinary: I mean use UTC time here (assuming the format is correct). There's a timezone offset below, so that should no longer with necessary:

```js
// Dates are in UTC; we need to do a timezone offset
const diff = now - dataSet[dataSet.length - 1].DT + (now.getTimezoneOffset() * 60000);
```

src/utils/ServerTime.js
MahboobehMohammadi: @brucebinary I wanted to reduce the server requests but 30 sec is okay too.

PR #528, opened 2018-09-20
set language via url GET param

brucebinary:


app/index.jsx
```diff
@@ -54,10 +54,11 @@ function getServerUrl() {
 const chartId = '1';
 const appId  = CIQ.localStorage.getItem('config.app_id') || 12812;
 const serverUrl = getServerUrl();
+const language = new URLSearchParams(window.location.search).get('l') || getLanguageStorage();
```
shayanbinary: @brucebinary I think `URLSearchParams` may not work on `Edge 15` and `Safari 9`, would you please check? if yes, what is your idea? change our support version or make custom function beside this one?

brucebinary: We're not supporting those right? [This is our browserlist setting](https://browserl.ist/?q=defaults%2C+not+op_mini+all%2C+not+ie+%3C%3D+11%2C+not+IE_Mob+%3C%3D+11%2C+not+Samsung+%3C%3D+4)

brucebinary: ok just talked with binary-static team. We need to support all edge versions.

brucebinary:
@shayanbinary I've updated the PR to use a polyfill if it is not supported.

shayanbinary:
@brucebinary I don't know if you fix this or not, but your test link has the problem on `Edge 16`

brucebinary:
@shayanbinary man. It just fails without any errors. But Edge 17 works. Damn.

brucebinary:
@shayanbinary omg. `URLSearchParams` is only added in edge 17. *_*

PR #533, opened 2018-09-26
Optimize the size of the po translations

brucebinary:
Translations account for 34% of the whole library. It is possible to shrink the library size.

This PR cuts down 109Kb (524-415) from the compiled library, or 10Kb (42-32) from the gzipped library. It achieves this by applying 2 optimizations:

 - Currently for each translation, the key that maps to the translation is the english translation. This is very inefficient. Therefore for all 13 languages, in the `smartcharts.js` there are 13 copies of the english translation. We can change this by accessing the translation by array index instead. In this implementation, a custom loader is added onto the `*.po` files to convert js objects to arrays.
 - Each translation has a `""` key that maps to information about the translation. This is useless for the actual translating process and can be removed.


MahboobehMohammadi:
LGTM

PR #542, opened 2018-10-05
add animation for sticky dialog on mobile

shayanbinary:


sass/components/_categorical-display.scss
```diff
@@ -374,20 +379,32 @@
...
+            .cq-lookup-panel {
+                position: relative;
+                overflow: hidden;
+
+                .cq-filter-panel {
+                    position: absolute;
+                    width: 100%;
+                    height: 50px;
+                    left: 0px;
+                    bottom: 0px;
```
shayanbinary: @brucebinary I put it here but it does not work. so I put it another place

PR #547, opened 2018-10-11
crosshair wrong Label

shayanbinary:


src/store/CrosshairStore.js
```diff
@@ -157,17 +157,24 @@ class CrosshairStore {
...
                 const rendererToDisplay = renderers[renderer];
                 const panel = stx.panels[rendererToDisplay.params.panel];
                 let { yAxis } = rendererToDisplay.params;
                 if (!yAxis && rendererToDisplay.params.shareYAxis) {
                     yAxis = panel.yAxis;
                 }
+
                 for (let id = 0; id < rendererToDisplay.seriesParams.length; id++) {
                     const seriesParams = rendererToDisplay.seriesParams[id];
+                    const symbolObject = symbolObjects.find(x => x.symbol === seriesParams.display);
```
brucebinary: I do not agree with this approach. We are doing a find operation for symbol objects _every_ single time the crosshair updates, which is a lot. The issue now after adding `display` to `addSeries` is simply that the crosshair does not show the correct display name. I think there is a better way to handle this.

shayanbinary:
@brucebinary amazing solution. it's great.

PR #548, opened 2018-10-11
fixing hiding scroll bar under fix header and auto-hie search bar in …

shayanbinary:
…mobile mode


MahboobehMohammadi:
LGTM

sass/components/_categorical-display.scss
```diff
@@ -361,6 +361,10 @@
             display: flex;
             justify-content: space-around;
             margin-bottom: -1px;
+            border-bottom: 1px solid #ff9900;
+            @include themify($themes) {
+                border-bottom: 1px solid themed('CatDisplaySearchBorder');
```
shayanbinary: @brucebinary this does not work here. how we should fix it, I think we should change the `themify` mixing

PR #559, opened 2018-10-21
add error handler for getTradingTimes api

shayanbinary:
**To replicate:**
Open chart on more than 3 tabs in a browser, and try to refresh these tabs 1 or 2 times

Raunak has an [answer](https://binary.vanillacommunity.com/discussion/328/you-have-reached-the-rate-limit-for-buy) for it.


brucebinary:
@shayanbinary this doesn't resolve the actual issue. It just accepts that there is an error in the response, outputs the error code, and go about its merry way without updating the trading times.

However, we do need more information on how to debug this. Let's keep that console error

PR #562, opened 2018-10-23
remove babel-polyfill and place it as peer dependency

brucebinary:
Since library users include babel-polyfill into their app, SmartChart should not need to do so:

**Before:**
`smartcharts.js   463 KiB       0  [emitted]  [big]  main`

**After:**
`smartcharts.js   388 KiB       0  [emitted]  [big]  main`

This results in 75Kb (16%) reduction of `smartcharts.js`


MahboobehMohammadi:
LGTM

PR #569, opened 2018-10-25
Mobile DOM

shayanbinary:


sass/styles/main.scss
```diff
@@ -123,6 +123,12 @@
     .ciq-menu-section {
         float: right;
     }
+    .smartcharts-mobile,
+    .smartcharts-desktop {
```
brucebinary: why do you need this?

shayanbinary: @brucebinary  we have `.smartcharts` in upper lines, but as we separate the these to classes in another `div` so to make sure everything work just like before, I inheritance these divs with their parent.

brucebinary: idk. The chart seems to work fine without it. I think you can remove this.

PR #579, opened 2018-11-05
Add Google Analytics

MahboobehMohammadi:


src/utils/ga.js
```diff
@@ -0,0 +1,24 @@
+import ReactGA from 'react-ga';
+
+export function initGA() {
```
brucebinary: for each of these functions (inside the exclude block), could you also add `window.location.host.endsWith('binary.com') || window.location.host.endsWith('binary.me')` to make sure the code only runs in `binary.[com|me]` domain. The `binary.sx` domains just adds noise the analytics data.

src/utils/ga.js
```diff
@@ -0,0 +1,24 @@
...
+    /* @START-EXCLUDE: 'lib' */
+    ReactGA.initialize('UA-128254050-1');
```
brucebinary: pls use binary's tag code: `UA-40877026-15`.

src/utils/ga.js
```diff
@@ -0,0 +1,24 @@
+import ReactGA from 'react-ga';
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary i look into it, if we cant remove it i try not to use it.

PR #582, opened 2018-11-08
Historical data

shayanbinary:


app/icons/ic-dropdown.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
brucebinary: you don't need this duplicate declaration. You should able to reuse this icon from the library

src/feed/Feed.js
```diff
@@ -32,7 +33,12 @@ class Feed {
     };
 
     onRangeChanged = () => {
-        console.log('RANGE CHANGE', this.startEpoch, this.endEpoch);
+        console.log('RANGE CHANGED', this.startEpoch, this.endEpoch);
+        const dtLeft =  new Date((this.startEpoch || this.endEpoch - ((this.granularity || 1) * 30)) * 1000);
```
brucebinary: I hard coded `30` here to just get it to work. Use `this._stx.chart.maxTicks` instead.

There is a piece of code in this file also using `(granularity || 1) * this._stx.chart.maxTicks`. See if you reuse that.

app/ChartHistory.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+        this.props.onChange(`${this.state.date} ${this.state.time}`);
+    }
+    render() {
+        return (
+            <div className="ciq-chart-history">
+                <DatePicker
+                    placeholder="select date"
+                    name="date"
+                    format="DD MMMM YYYY"
```
shayanbinary: @brucebinary Should we read this `format` from the application?

src/components/ChartSetting.jsx
```diff
@@ -94,6 +96,15 @@ const ChartSetting = ({
                         </div>
                     </div>
                     {!isMobile ? <AssetInformationToggle value={assetInformation} onChange={setAssetInformation} /> : ''}
+                    <div className="ciq-list-item">
+                        <span className="ciq-icon-text">{t.translate('Historical Data Mode')}</span>
```
brucebinary: you need to run `yarn translations`

index.html
```diff
@@ -7,7 +7,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
     <meta name="description" content="Analyse the financial markets using beautiful visualisations and unique trading tools, powered by a fast and intuitive interface." />
     <title>SmartCharts | Binary.com</title>
-    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css" media="screen" />
+    <link rel="stylesheet" type="text/css" href="./dist/smartcharts.css?ceeb9b00" media="screen" />
```
brucebinary: don't commit the hashes

app/ChartHistory.jsx
brucebinary: @shayanbinary did I answer you on this on slack? I don't understand your question.

app/ChartHistory.jsx
```diff
@@ -0,0 +1,56 @@
...
+                time: target.value,
+            });
+        }
+
+        this.props.onChange(`${this.state.date} ${this.state.time}`);
+    }
+    render() {
+        return (
+            <div className="ciq-chart-history">
+                <strong>Historical Data:</strong>
```
brucebinary: this needs to be translated as well. I think for the time being we can just bundle the app and library translations together (there are not many strings in the app anyway)

src/components/ChartControls.jsx
```diff
@@ -12,13 +12,13 @@ import Share from './Share.jsx';
...
 
-export const RenderDefaultControls = ({ isMobile }) => (
+export const RenderDefaultControls = ({ isMobile, isHistorical }) => (
     <>
         {isMobile ? '' : <CrosshairToggle />}
         <ChartTypes />
         <Timeperiod />
         <StudyLegend />
-        <Comparison />
+        {isHistorical ? '' : <Comparison />}
```
brucebinary: is this suppose to disable comparisons when historical mode is on? It's not doing so. The controls are rendered in the app code. In fact, you should not put this logic in the library since the library doesn't know there's something called historical mode. It's just a switch in the settings.

src/components/Icons.jsx
```diff
@@ -169,6 +172,9 @@ export const SettingIcon = Wrapper(Edit);
 export const ArrowIcon = Wrapper(Arrow);
 export const FavoriteIcon = Wrapper(Star);
 export const DownloadIcon = Wrapper(Download);
+export const CalendarIcon = Wrapper(Calendar);
+export const TimeIcon = Wrapper(Time);
+export const CloseCircleIcon = Wrapper(CloseCircle);
```
brucebinary: we only want these 3 icons in the app. What you can do is export out `Wrapper` instead (it could use a better name), and import those svg in the app.

src/store/ChartSettingStore.js
```diff
@@ -11,7 +11,7 @@ export default class ChartSettingStore {
         this.mainStore = mainStore;
         this.menu = new MenuStore(mainStore, { route: 'setting' });
         this.ChartSettingMenu = this.menu.connect(Menu);
-        reaction(() => mainStore.state.settings, () => {
+        reaction(() => this.context, () => {
```
brucebinary: why did you change this?

src/store/ChartStore.js
```diff
@@ -519,6 +519,14 @@ class ChartStore {
             this.stxx = null;
         }
     }
+
+    @action.bound onHistorical() {
+        this.contextPromise.then(() => {
+            // console.log('asdas');
```
brucebinary: dude. Please check your code (every single line) before you send for review

src/store/ChartStore.js
```diff
@@ -519,6 +519,14 @@ class ChartStore {
...
+            this.mainStore.chartType.setType('mountain');
+            this.changeSymbol(undefined, 0);
```
brucebinary: So I understand you added `onHistorical` so that when the user selects historical mode you change to a line chart with ticks granularity. 

A better way to do this is from the app. When the app gets the settings that it is in historical mode, it sets these exact same settings using props

app/ChartHistory.jsx
shayanbinary: @brucebinary actually no, do date format (`DD MMMM YYYY`) comes from any setting?

src/store/ChartSettingStore.js
shayanbinary: oh, so sorry, perhaps I am trying to test something and forget to clean it

src/store/ChartStore.js
shayanbinary: so sorry

src/store/ChartStore.js
```diff
@@ -282,6 +284,14 @@ class ChartStore {
         const studiesStore = this.mainStore.studies;
         stxx.callbacks.studyOverlayEdit = studiesStore.editStudy;
         stxx.callbacks.studyPanelEdit = studiesStore.editStudy;
+        const chartiqHome = stxx.home;
+        stxx.home = function () {
+            if (settings.historical) {
+                feed.onRangeChanged();
+            } else {
+                chartiqHome.apply(this, {});
```
shayanbinary: @brucebinary what I come to at last I tried to test `setRange`, but not work properly it didn't show correct range, you should scroll a little after setRange called, so I use feed range changed

brucebinary: Consider using endEpoch value instead of historical setting. Instead of using setRange you can set scroll. Or both. 

You could just call chartiqHome().

src/components/ui/Animation.js
```diff
@@ -65,6 +65,7 @@ export default function animateChart(stx, animationParameters, easeMachine) {
         if (params !== undefined && params.animationEntry) return;
 
         function completeLastBar(value) {
+            if (!chart.masterData) { return; }
```
shayanbinary: @brucebinary when switching to historical data `masterData` get null, so that `.length`  throw an error, I put this piece of code for this reason.

app/index.jsx
```diff
@@ -151,7 +164,8 @@ class App extends Component {
             window.location.href = `${origin}${pathname}?${url.toString()}`;
         }
         if (settings.historical) {
-            this.handleDateChange(today);
+            this.cleanComparison();
+            window.location.reload(false);
```
shayanbinary: while user working with the chart, we should remove added `comparisons`, so we just try to update local storage and then reload the page so that chart load these data again.

brucebinary: yes, it makes sense to refresh the app - since there's no way to programmatically set the comparisons.

app/index.jsx
```diff
@@ -124,6 +138,17 @@ class App extends Component {
...
     }
     */
+    cleanComparison = () => {
+        try {
+            const layout_string = localStorage.getItem(`layout-${chartId}`),
+                layout = JSON.parse(layout_string !== '' ? layout_string : '{}');
+
+            layout.symbols.splice(1, layout.symbols.length - 1);
+            localStorage.setItem(`layout-${chartId}`, JSON.stringify(layout));
+        } catch (e) {
```
brucebinary: why do you need this silent try catch block here?

also I think `removeAllComparisons` is a better name for this method.

src/feed/Feed.js
```diff
@@ -177,6 +176,9 @@ class Feed {
         callback({ quotes });
 
         this._emitDataUpdate(quotes, comparisonChartSymbol);
+        if (this.endEpoch) {
+            this._stx.home();
```
shayanbinary: to make sure, on `historical data` graph always scroll to right side on each time select

src/store/ChartStore.js
```diff
@@ -118,6 +120,137 @@ class ChartStore {
...
+                while (this.micropixels < 0) {
+                    exactScroll--;
+                    this.micropixels += layout.candleWidth;
+                }
+                this.micropixels -= layout.candleWidth;
+                exactScroll++;
+                if ((!this.mainSeriesRenderer || !this.mainSeriesRenderer.standaloneBars) && !this.standaloneBars[layout.chartType]) this.micropixels += layout.candleWidth / 2; // bar charts display at beginning of candle
+
+                if (this.isHistoricalMode() && _self.isMobile) {
+                    exactScroll = parseInt(exactScroll * 0.8, 10);
```
shayanbinary: as there are a label added to last spot, so scroll should be reduce to show all label text

app/DateTimePicker/Calendar.jsx
```diff
@@ -0,0 +1,291 @@
...
+    const is_year_view   = calendar_view === 'year';
+    const is_decade_view = calendar_view === 'decade';
+    const moment_date    = moment.utc(calendar_date);
+
+    return (
+        <div className="calendar-header">
+            <CalendarButton
+                className={`calendar-prev-year-btn ${isPeriodDisabled(moment_date.clone().subtract(1, 'month'), 'month') ? 'hidden' : ''}`}
+                onClick={() => (
+                    (is_date_view || is_month_view) && onClick.previousYear())
```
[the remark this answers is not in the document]
shayanbinary: @morteza-binary Actually I just grab this code from binary static, and just did some small modifications on it.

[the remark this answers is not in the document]
morteza-binary: If you mean binary2.0, also we have changed the code there.

sass/styles/_menu.scss
```diff
@@ -107,7 +107,7 @@ $transition-duration: 300ms;
     .cq-modal-dropdown {
         border: none;
         pointer-events: none;
-        z-index: 1;
+        z-index: -1;
```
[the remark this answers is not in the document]
shayanbinary: @morteza-binary yes, to fix the clicking issue on mobile devices

chartiq/chartiq.js
```diff
@@ -5186,7 +5186,7 @@ var __js_span_ =
 				self.setCandleWidth(previousCandleWidth);
 				self.layout.interval =  previousInterval;
 				self.layout.periodicity = previousPeriodicity;
-				self.layout.timeUnit= previousTimeUnit;
+				self.layout.timeUnit = previousTimeUnit;
```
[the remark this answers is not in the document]
shayanbinary: perhaps that was my mistake

package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
```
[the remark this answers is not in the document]
shayanbinary: oops

app/DateTimePicker/panels/calendar_months.jsx
```diff
@@ -0,0 +1,39 @@
...
+    return (
+        <div className="calendar-month-panel">
+            {Object.keys(month_headers).map((month, idx) => (
+                <span
+                    key={month}
+                    className={`calendar-month ${(idx === selected_month) ? 'active' : ''} ${isPeriodDisabled(moment_date.month(month), 'month') ? 'disabled' : ''} `}
+                    onClick={onClick.month}
+                    data-month={idx}
+                >
+                    {t.translate(month_headers[month])}
```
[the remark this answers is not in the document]
shayanbinary: @morteza-binary Oops you are right, my bad. So any solution or Idea about this?

app/index.jsx
```diff
@@ -124,6 +169,17 @@ class App extends Component {
...
     */
+    removeAllComparisons = () => {
+        try {
+            const layoutString = localStorage.getItem(`layout-${chartId}`),
+                layout = JSON.parse(layoutString !== '' ? layoutString : '{}');
+
+            layout.symbols.splice(1, layout.symbols.length - 1);
+            localStorage.setItem(`layout-${chartId}`, JSON.stringify(layout));
+        } catch (e) {
+            console.log(e);
```
MahboobehMohammadi: @shayanbinary do you need this here?

chartiq/chartiq.js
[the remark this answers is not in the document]
MahboobehMohammadi: @shayanbinary did you revert this?

src/store/CurrentSpotStore.js
```diff
@@ -47,10 +51,17 @@ class CurrectSpotStore {
                 visible = false;
             }
         }
-
-        this.show = visible && (layout.chartType !== 'candle'
-            && layout.chartType !== 'colored_bar'
-            && layout.chartType !== 'hollow_candle');
+        this.historical = this.state.endEpoch
```
shayanbinary: these codes are for converting current spot to the historical line

PR #586, opened 2018-11-09
Refactor ga

MahboobehMohammadi:


src/utils/ga.js
```diff
@@ -1,29 +1,54 @@
...
 export function logEvent(category, action, label) {
     /* @START-EXCLUDE: 'lib' */
-    if (!window.isProductionWebsite()) return;
-    ReactGA.event({
-        category,
-        action,
-        label,
-    });
+    if (!window.isProductionWebsite) return;
+    try {
```
brucebinary: instead of a try catch block you can set `ga` implementation to be just a blank function.

src/utils/ga.js
```diff
@@ -1,29 +1,54 @@
...
+        ga('send', 'event', category, action, label);
```
brucebinary: I think it's better to place these ga code as part of the app. We can set smartcharts to use a default ga function (which is blank), granted `ga` has not already been defined.

src/utils/ga.js
MahboobehMohammadi: @brucebinary its just a precaution, i imported analytics.js before ga is being used. but since its being imported asynchronously, i add try catch. why do we need a back function?

PR #589, opened 2018-11-14
Add table chart type

MahboobehMohammadi:


src/store/ChartTableStore.js
```diff
@@ -0,0 +1,84 @@
...
+        if (this.open) {
+            this.stx.showTable = true;
+            this.updateTableData(this.stx.masterData);
+        } else {
+            this.stx.showTable = false;
+            this.tableData =  [];
+        }
+    }
+
+    @action.bound updateTableData(ticks) {
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary they are two different function, return different table data.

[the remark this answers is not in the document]
brucebinary: @MahboobehMohammadi hmm. Thought there were some similarities between the 2 functions, but ok.

src/store/ChartTableStore.js
```diff
@@ -45,9 +45,9 @@ export default class ChartTableStore {
         const year = DT.getUTCFullYear();
         const month = DT.getUTCMonth() + 1;
         const day = DT.getUTCDate();
-        const hours = DT.getUTCHours();
-        const minutes = DT.getUTCMinutes();
-        const seconds = DT.getUTCSeconds();
+        const hours = DT.getHours();
+        const minutes = DT.getMinutes();
+        const seconds = DT.getSeconds();
```
brucebinary: ...why did you change this?

All of our data should be in UTC time.

MahboobehMohammadi: @brucebinary it is already in UTC time, it was double converting it!

brucebinary: @MahboobehMohammadi omg surprised our clients didn't spot that. :O

Thanks for spotting this.

if it is already in UTC time, shouldn't the year month and day also be already in UTC time?

MahboobehMohammadi: @brucebinary yes right, but arent they the same? i mean when does UTC year could be different from local year?! or month.

brucebinary: @MahboobehMohammadi At the end of each year. Different parts of the world reach the new year at different times.

src/store/ChartTableStore.js
```diff
@@ -42,8 +42,8 @@ export default class ChartTableStore {
     @action.bound updateTableData({ DT, Open, High, Low, Close }) {
         this.isTick = this.mainStore.timeperiod.timeUnit === 'tick';
 
-        const year = DT.getUTCFullYear();
-        const month = DT.getUTCMonth() + 1;
+        const year = DT.getFullYear();
+        const month = DT.getMonth() + 1;
         const day = DT.getUTCDate();
```
brucebinary: as mentioned earlier, this should also be `getDate()` because it is already in UTC time

src/store/ShareStore.js
```diff
@@ -62,8 +62,8 @@ export default class ShareStore {
         const header = `Date,Time,${isTick ? this.marketDisplayName : 'Open,High,Low,Close'}`;
         const lines = [header];
         this.stx.masterData.forEach(({ DT, Open, High, Low, Close }) => {
-            const year = DT.getUTCFullYear();
-            const month = DT.getUTCMonth() + 1; // months from 1-12
+            const year = DT.getFullYear();
+            const month = DT.getMonth() + 1; // months from 1-12
             const day = DT.getUTCDate();
```
brucebinary: ^ same as comment above

PR #592, opened 2018-11-21
adding clear all button to draw tool list

shayanbinary:


MahboobehMohammadi:
LGTM.

src/store/DrawToolsStore.js
```diff
@@ -12,6 +10,41 @@ const formatCamelCase = (s) => {
     const capitalized = s.charAt(0).toUpperCase() + s.slice(1);
     return capitalized.replace(/([a-z](?=[A-Z]))/g, '$1 ');
 };
+const DrawToolsItems = [
```
brucebinary: this will fail to translate the strings here when we change language.

brucebinary: @shayanbinary this is cause the strings here not to be added to `messages.pot` when you run `yarn translations`

MahboobehMohammadi:
LGTM

PR #593, opened 2018-11-22
Refreshing chart on screen lock

shayanbinary:


shayanbinary:
with [`Idle-js`](https://www.npmjs.com/package/idle-js) we are able to detect if a user enters/leave the current tab, or be idle or active, so we can call `refreshChart` whenever user get active or enter the tab, but I don't think this would be a good solution. So please check my second commit on `data breathing`

PR #594, opened 2018-11-27
upgrade chartIQ to version v6.2.2

shayanbinary:


src/store/ChartStore.js
```diff
@@ -446,6 +446,7 @@ class ChartStore {
     setMainSeriesDisplay(name) {
         // Set display name of main series (to be shown in crosshair tooltip)
         this.stxx.chart.seriesRenderers._main_series.seriesParams[0].display = name;
+        this.stxx.chart.seriesRenderers._main_series.seriesParams[0].field = 'Close';
     }
```
shayanbinary: @brucebinary After some hours working on this, and according to ChartIQ Team, I use this temporary solution for crosshair

brucebinary: @shayanbinary as long as it works. >_< if the solution is temporary, do add a `//TODO...` to indicate that it needs to be removed in the future.

src/store/ChartStore.js
```diff
@@ -446,6 +429,7 @@ class ChartStore {
     setMainSeriesDisplay(name) {
         // Set display name of main series (to be shown in crosshair tooltip)
         this.stxx.chart.seriesRenderers._main_series.seriesParams[0].display = name;
+        // TODO
```
brucebinary: @shayanbinary what's this.

shayanbinary: @brucebinary :D a todo so that we remember, we should change this line of code

PR #598, opened 2018-12-03
Fix crosshair tooptip position

MahboobehMohammadi:


src/store/CrosshairStore.js
```diff
@@ -316,7 +316,7 @@ class CrosshairStore {
     updateTooltipPosition() {
         this.isArrowLeft = CIQ.ChartEngine.crosshairX <= MAX_TOOLTIP_WIDTH;
         this.left = CIQ.ChartEngine.crosshairX;
-        this.top = CIQ.ChartEngine.crosshairY;
+        this.top = CIQ.ChartEngine.crosshairY - this.stx.top;
```
brucebinary: @MahboobehMohammadi why is this needed now? The crosshair position looks fine in beta

MahboobehMohammadi: @brucebinary as you see here crosshair position is not correct when chart is not at the top left of screan since `CIQ.ChartEngine.crosshairX` returns current X screen coordinate of the crosshair.

MahboobehMohammadi: @brucebinary ?

brucebinary: @MahboobehMohammadi ops. Lost track of this. Do make a card for this so it can be tracked from the board.

brucebinary: @MahboobehMohammadi owh ok I saw the PR 🤦‍♂️

PR #599, opened 2018-12-04
Buffer pending requests

MahboobehMohammadi:


app/connection/ConnectionManager.js
```diff
@@ -85,7 +87,10 @@ class ConnectionManager extends EventEmitter {
         }
 
         Object.keys(this._pendingRequests).forEach((req_id) => {
-            this._pendingRequests[req_id].reject('Pending requests are rejected as connection is closed.');
+            const data = this._pendingRequests[req_id].data;
```
brucebinary: nitpick: consider using destructuring

app/connection/ConnectionManager.js
```diff
@@ -110,6 +115,13 @@ class ConnectionManager extends EventEmitter {
         }, timeout);
     }
 
+    _sendBufferedRequests() {
```
brucebinary: can you add a `console.error` message here? I would like to know whether this method would ever get called, since it only appears in TrackJS. Be sure to indicate it's not an error.

MahboobehMohammadi: @brucebinary it happens in a very rare case. you have to go offline when chart is not loaded yet.
image

src/store/ChartStore.js
```diff
@@ -445,6 +445,8 @@ class ChartStore {
 
     // Calling newChart with symbolObj as undefined refreshes the chart
     @action.bound newChart(symbolObj = this.currentActiveSymbol, params) {
+        if(!symbolObj)
```
brucebinary: this changes the logic altogether. Setting `symbolObj` to undefined should cause the chart to refresh, not do nothing.

actually I think there's a bug with `refreshChart`. The implementation should be `this.newChart(undefined)`.

app/connection/ConnectionManager.js
```diff
@@ -85,7 +85,7 @@ class ConnectionManager extends EventEmitter {
         }
 
         Object.keys(this._pendingRequests).forEach((req_id) => {
-            this._pendingRequests[req_id].reject('Pending requests are rejected as connection is closed.');
+            console.log(`Pending requests are rejected as connection is closed. Request Id= ${req_id}`);
```
brucebinary: This leaves the promise unresolved for the lifetime of the promise, which meant that the state is `PENDING`, which is not - we are discarding the promise (which we don't want to do - we want to continue to retry when possible). Better that you catch the error than silently hide it.

src/feed/Feed.js
```diff
@@ -331,7 +331,10 @@ class Feed {
 
     _onConnectionReopened() {
         const keys = Object.keys(this._activeStreams);
-        if (keys.length === 0) { return; }
+        if (keys.length === 0) { 
+            this._mainStore.chart.refreshChart();
```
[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary when connection closes we send forgetrequests and remove all the streams so we need to create a new stream

src/store/ChartStore.js
MahboobehMohammadi: @brucebinary im not sure, but as far as i understood if `symbolObj` is `undefind` we are using `this.currentActiveSymbol` we do nothing which means the `chart`  is not completly initialized so we dont need realy to refresh, i needed  `refreshchart` only to because when connection closes we forget all the streams so this only is realy needed there which `currentActiveSymbol` has value so it refresh the chart.

app/connection/ConnectionManager.js
MahboobehMohammadi: @brucebinary right, i shouldnt have create a new promise, i send the new req with current promise so it will resolve it when connection back, right?

brucebinary: send the existing request, reusing the same promise.

app/connection/ConnectionManager.js
```diff
@@ -110,7 +116,15 @@ class ConnectionManager extends EventEmitter {
         }, timeout);
     }
 
-    async send(data, timeout) {
+    _sendBufferedRequests() {
+        while (this._bufferedRequests.length > 0) {
+            const req = this._bufferedRequests.shift();
+            this.send( req.data , req);
```
brucebinary: You changed the method signature to `async send(data, promise, timeout) {`? This doesn't match

app/connection/ConnectionManager.js
```diff
@@ -122,7 +136,7 @@ class ConnectionManager extends EventEmitter {
         }
 
         this._websocket.send(JSON.stringify(req));
-        this._pendingRequests[req.req_id] = new PendingPromise(req);
+        this._pendingRequests[req.req_id] = promise ? promise: new PendingPromise(req);
```
brucebinary: I think it's better to have `req_id` passed to `send` as an optional parameter. For this part, instead of using a passed promise, check if the pending request exist inside the object and reuse that instead

src/feed/Feed.js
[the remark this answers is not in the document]
brucebinary: when the connection closes, forget requests are not sent (see https://github.com/binary-com/SmartCharts/blob/dev/src/feed/subscription/RealtimeSubscription.js#L9). Streams are not removed but paused.

The only instance (at least as I recall) where we want refresh chart to be used is when we are too far behind tick history, so just refreshing is preferable than getting a large ticks_history response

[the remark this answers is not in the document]
MahboobehMohammadi: @brucebinary Check `_onConnectionClosed` in `streamManager` 
https://github.com/binary-com/SmartCharts/blob/27f72816f0be503e6cf5190d3c5c83271d2f63be/app/connection/StreamManager.js#L44

app/connection/ConnectionManager.js
```diff
@@ -59,14 +60,16 @@ class ConnectionManager extends EventEmitter {
...
+
         if (!this._pingTimer) {
             this._pingTimer = setInterval(this._pingCheck.bind(this), 15000);
         }
     }
 
     _pingCheck() {
         if (this._websocket.readyState === WebSocket.OPEN) {
-            this.send({ ping: 1 }, 5000)
+            this.send({ ping: 1 }, undefined, 5000)
```
brucebinary: don't forget this ^

app/connection/ConnectionManager.js
brucebinary: @MahboobehMohammadi The log message does not follow what the code is actually doing now.

app/connection/StreamManager.js
```diff
@@ -41,7 +41,9 @@ class StreamManager {
         // It is not its responsibility to reestablish the streams upon reconnection.
         this._streamIds = {}; // set it to blank so that forget requests do not get called
         for (const key of Object.keys(this._streams)) {
-            this._forgetStream(key);
+            if (this._streams[key].subscriberCount !== 0) {
```
brucebinary: we shouldn't need this. Forget requests do not get sent when `this._streamIds` is set to `{}`. The code block here simply cleans up the unused streams from memory.

brucebinary: I was able to get it working without `if (this._streams[key].subscriberCount !== 0) { }`

MahboobehMohammadi: @MahboobehMohammadi the forgetrequest is beign send here. cause we removed the stream.
https://github.com/binary-com/SmartCharts/blob/1003dfe3861a7f785691b137e024c348e0ba44a7/app/connection/StreamManager.js#L25

PR #614, opened 2018-12-26
Fix chart stopping after screen lock

MahboobehMohammadi:


app/connection/robust-websocket.js
```diff
@@ -0,0 +1,246 @@
...
+            clearTimeout(connectTimeout)
+            connectTimeout = null
+          }
+          event.reconnects = ++reconnects
+          event.attempts = attempts
+          attempts = 0
+          reconnectWhenOnlineAgain = false
+        }],
+        close: [function(event){
+          setTimeout(() => {
```
MahboobehMohammadi: @brucebinary here when the mobile phone wifi goes offline, the ws connection closes before browser detect the connection change so `navigator.online` is still true when it tries to reconnect.

PR #616, opened 2018-12-27
upgrade the resize-observer-polyfill

shayanbinary:


package.json
```diff
@@ -120,7 +120,7 @@
   "dependencies": {
     "event-emitter-es6": "^1.1.5",
     "lodash.debounce": "^4.0.8",
-    "resize-observer-polyfill": "^1.5.0",
+    "robust-websocket": "^0.3.0",
```
brucebinary: we don't use this anymore

PR #621, opened 2019-01-01
Add supportAnimation prop to smartchart

Mahboobeh-binary:


src/store/ChartStore.js
```diff
@@ -147,6 +148,7 @@ class ChartStore {
         chartSetting.onSettingsChange = onSettingsChange;
         this.isMobile = isMobile;
         this.state = this.mainStore.state;
+        this.supportsAnimation = supportsAnimation;
```
brucebinary: this should be in `ChartState`.

 I think it is safe to assume that library user would set this once and not go back and forth with this setting - therefore we can stop `animateChart(stxx, { stayPut: true });` and `this.stx.append('draw', this.updateSpot);` from bring called altogether.

Also do add this setting to documentation + explain why we want to disable it.

MahboobehMohammadi: @brucebinary yes this also will work if we wanna disable it all together, though i think its better to follow their documentation https://documentation.chartiq.com/CIQ.Animation.html#main.

brucebinary: > Keep in mind that changing to a different chart type, may once again enable animation. You can override this by adding an event listener on layout changes.

...that's too much hassle - considering our use case.

MahboobehMohammadi: @brucebinary sure. i already used your approach!

src/store/ChartState.js
```diff
@@ -23,7 +23,7 @@ class ChartState {
         this.stxx.addEventListener('drawing', this.saveDrawings.bind(this));
     };
 
-    @action.bound updateProps({ id, settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch }) {
+    @action.bound updateProps({ id, settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch, supportsAnimation }) {
```
brucebinary: For consistency with the rest of the API, we should call it `isAnimationEnabled`

we can set default value of in this line:  `...endEpoch, isAnimationEnabled = true }`

PR #625, opened 2019-01-04
Use simplebar-react instead of simple bar due to lib errors

MahboobehMohammadi:


src/components/categoricaldisplay/CategoricalDisplay.jsx
```diff
@@ -1,27 +1,38 @@
...
-            className="cq-scroll-panel"
-            ref={setScrollPanel}
-        >
-            <ResultsPanel
-                onSelectItem={onSelectItem}
-            />
-        </div>
-    </div>
-);
+class CategoricalDisplay extends React.Component {
```
brucebinary: why do you need to make this a component?

MahboobehMohammadi: @brucebinary i dont. just revert it.

PR #629, opened 2019-01-10
set new colors for price label

shayanbinary:


src/store/ChartStore.js
```diff
@@ -123,6 +123,59 @@ class ChartStore {
...
+            let yaxisLabelStyle = this.yaxisLabelStyle;
+            if (yax.yaxisLabelStyle) yaxisLabelStyle = yax.yaxisLabelStyle;
+            const params = {
+                ctx:context,
+                x,
+                y,
+                top:y - (height / 2),
+                width,
+                height,
+                radius,
```
shayanbinary: Referring to the design of `current price label` it shouldn't have a radius, and this prop only can be handled here. that's why I use monkey patching for here.

src/store/CurrentSpotStore.js
```diff
@@ -47,9 +47,7 @@ class CurrectSpotStore {
             }
         }
 
-        this.show = visible && (layout.chartType !== 'candle'
-                && layout.chartType !== 'colored_bar'
-                && layout.chartType !== 'hollow_candle');
+        this.show = visible && (layout.aggregationType !== 'pandf');
```
shayanbinary: As Designers request to show spot on every chart type except `point & figure`.

src/store/ChartStore.js
```diff
@@ -123,6 +123,59 @@ class ChartStore {
         this.chartNode = this.rootNode.querySelector('.ciq-chart-area');
         this.chartControlsNode = this.rootNode.querySelector('.cq-chart-controls');
 
+        CIQ.ChartEngine.prototype.createYAxisLabel = function (panel, txt, y, backgroundColor, color, ctx, yAxis) {
```
MahboobehMohammadi: @shayanbinary I dont think its good to override the whole `createYAxisLabel `, if we cant change the style can you check for a listener or injecttion, something like `headsUpHR` and just change float label inline style there?

shayanbinary: @MahboobehMohammadi actually by `drawCurrentHR` you just be able to set color and background, but for the new design, we need to handle `radius, height and width` which these props effect directly from this method

PR #630, opened 2019-01-10
Replace simplebar with tt-react-custom-scrollbar

MahboobehMohammadi:


src/store/CategoricalDisplayStore.js
```diff
@@ -272,11 +272,12 @@ export default class CategoricalDisplayStore {
     }
 
     @action.bound setScrollPanel(element) {
-        this.scrollPanel =  element.querySelector('.simplebar-content');
+        this.scrollPanel =  element;
```
shayanbinary: @MahboobehMohammadi `element` maybe `null` in some case, please handle it also

src/store/CategoricalDisplayStore.js
```diff
@@ -285,5 +286,6 @@ export default class CategoricalDisplayStore {
         ResultsPanel: this.ResultsPanel,
         FilterPanel: this.FilterPanel,
         SearchInput: this.SearchInput,
+        isMobile: this.mainStore.chart.isMobile,
```
shayanbinary: maybe it's better to get chart as a `getter` like `get chart(){return this.mainStore.chart}`

src/components/DrawTools.jsx
```diff
@@ -24,7 +24,10 @@ const DrawTools = ({
 
         <DrawToolsMenu.Body>
             <div className="body">
-                <SimpleBar
+                <Scrollbars
+                    autoHide
+                    autoHeight
+                    autoHeightMax={260}
```
shayanbinary: @MahboobehMohammadi  perhaps it's better to handle `max-height` in the CSS

src/components/PriceLine.jsx
```diff
@@ -26,9 +26,9 @@ class PriceLine extends Component {
                 className={`chart-line horizontal ${draggable ? 'draggable' : ''} ${isDragging ? 'dragging' : ''} ${className || ''}`}
                 style={{ top, zIndex, color: draggable ? color : '#FFF', borderColor: color }}
                 ref={setDragLine}
-                hidden={visible ? undefined : 'true'}
-                uncentered={uncentered ? 'true' : undefined}
-                off-screen={offScreen ? 'true' : undefined}
+                hidden={visible ? undefined : true}
+                uncentered={uncentered ? true : undefined}
+                off-screen={offScreen ? true : undefined}
```
kellybinary: i think this can be simplified to `true || undefined`

src/store/CategoricalDisplayStore.js
MahboobehMohammadi: @shayanbinary in which cases it might be null?

PR #638, opened 2019-01-22
on changing symbol, assert that comparison series are removed from Ch…

brucebinary:
…artIQ before creating new chart

### DIFF
--- a/src/store/ChartStore.js
+++ b/src/store/ChartStore.js
@@ -419,6 +419,14 @@ class ChartStore {
             params = { periodicity: calculateTimeUnitInterval(granularity) };
         }
 
+        if (params === undefined && symbolObj) {
+            for (const field in this.stxx.chart.series) {
+                if (this.stxx.chart.series[field].parameters.bucket !== 'study') {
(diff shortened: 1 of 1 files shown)


brucebinary:
@MahboobehMohammadi you can close this PR when you're done with it. It's never intended to be a final solution

src/store/ChartStore.js
```diff
@@ -419,6 +419,14 @@ class ChartStore {
             params = { periodicity: calculateTimeUnitInterval(granularity) };
         }
 
+        if (params === undefined && symbolObj) {
+            for (const field in this.stxx.chart.series) {
+                if (this.stxx.chart.series[field].parameters.bucket !== 'study') {
```
MahboobehMohammadi: @brucebinary It looks fine and works fine, thanks! except I don't think chartiq add series for indicators

brucebinary: @MahboobehMohammadi this line is from their original source code some versions back. It could be outdated though.

morteza-binary:
Duplicate of #650

PR #640, opened 2019-01-24
Chart Overall Look

ashkanx:


sass/_themes.scss
```diff
@@ -1,294 +1,295 @@
...
-// Dark Theme Colors
-$COLOR_BRANDBLUE_5 : #f4f4f6;
-$COLOR_BRANDBLUE_40: #aaacba;
-$COLOR_BRANDBLUE_75: #60647e;
-$COLOR_BRANDBLUE_85: #4a4f6c;
-$COLOR_BRANDBLUE_95: #353b5b;
+$COLOR_WHITE        : #ffffff;
+$COLOR_BLACK        : #000000;
+$COLOR_LIGHT_BLACK_1: rgba($COLOR_BLACK, 0.8);
+$COLOR_LIGHT_BLACK_2: rgba($COLOR_BLACK, 0.4);
```
MahboobehMohammadi: @ashkanx i coulnt find where do we use this `COLOR_LIGHT_BLUE_1` to `COLOR_LIGHT_BLUE_5`

ashkanx: @MahboobehMohammadi I added the all colours according to Zeplin projects for both light and dark themes. so these are used later

sass/components/_timeperiod.scss
```diff
@@ -195,7 +195,7 @@
         top: 0px;
 
         > .cq-menu-btn {
-            padding: 16px 5px !important;
+            // padding: 16px 5px !important;
```
MahboobehMohammadi: @ashkanx if the padding is not needed we can just remove it  and please remove it fot ` .smartcharts-480 ` as well. its in line 190

PR #642, opened 2019-01-28
Add last digit history

MahboobehMohammadi:


src/store/LastDigitsStore.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @action.bound async showLastDigitStats() {
+        if (!this.context) return;
+        if (this.mainStore.state.showLastDigitStats) {
+            for (let i = 0; i < 10; i++) {
+                this.digits.push(0);
+                this.bars.push({ height:0, cName:'' });
+            }
+            const tickHistory = await this.api.getTickHistory({ symbol :this.mainStore.chart.currentActiveSymbol.symbol });
```
morteza-binary: @MahboobehMohammadi I'd be better to define the `1000` in the request here instead of using `BinaryAPI.DEFAULT_COUNT`. If in the future because of any reason the default value is changed, it won't affect the functionality of the component.

Indeed, we can use the previous response of `ticks_history` instead of making another request. If they are not as many as 1000 ticks then we can send a request to get the latest 1000.

src/store/LastDigitsStore.js
```diff
@@ -0,0 +1,86 @@
+import { observable, action, computed, reaction } from 'mobx';
+
+export default class LastDigitsStore {
+    constructor(mainStore) {
+        this.mainStore = mainStore;
+        reaction(() => this.mainStore.chart.currentActiveSymbol, this.showLastDigitStats);
```
morteza-binary: @MahboobehMohammadi it'd be better to use `componentDidUpdate` lifecycle method in `LastDigits` component and check the `marketDisplayName` property to reload the data.
```
componentDidUpdate({ marketDisplayName }) {
    if (this.props.marketDisplayName !== marketDisplayName) {
        this.props.showLastDigitStats();
    }
}
```

MahboobehMohammadi: @morteza-binary what is the difference?

morteza-binary: @MahboobehMohammadi It's more readable for other developers and makes the maintenance much easier. 
Also in this case, if you need to use reaction you should clear reaction when the component is not mounted and create it after mounting.

ashkanx:
@MahboobehMohammadi conflicts

PR #650, opened 2019-02-08
Comparison issue

MahboobehMohammadi:


sass/components/_timeperiod.scss
```diff
@@ -187,7 +187,7 @@
         top: 0px;
 
         > .cq-menu-btn {
-            padding: 16px 5px !important;
+            //padding: 16px 5px !important;
```
morteza-binary: @MahboobehMohammadi Would you plz remove the commented line and even the class name since it's not usable anymore.

ashkanx: That was me :|

MahboobehMohammadi: @morteza-binary yes ashkan is working on this so i kept it so he knows what changes, he can remove comments after his done.

src/store/ChartStore.js
```diff
@@ -552,6 +552,13 @@ class ChartStore {
             params = { periodicity: calculateTimeUnitInterval(granularity) };
         }
 
+        if (params === undefined && symbolObj) {
+            // Remove comparisons before symbol changes
+            for (const field in this.stxx.chart.series) {
+                this.stxx.removeSeries(field);
```
morteza-binary: @MahboobehMohammadi should we clear `_streams` and `_tickHistoryCache` after removing the series from the chart to prevent the cache from being too fat?

MahboobehMohammadi: @morteza-binary no.its been handled somewhere elese

PR #653, opened 2019-02-09
manually change the color of grid lines

shayanbinary:


src/store/ChartSettingStore.js
```diff
@@ -119,7 +120,12 @@ export default class ChartSettingStore {
     @action.bound setTheme(theme) {
         if (this.theme === theme) { return; }
         this.theme = theme;
-        if (this.context) { this.stx.clearStyles(); }
+
+        if (this.context) {
+            this.stx.clearStyles();
+            this.stx.setStyle('stx_grid', 'color', Theme[`${theme}chartgrid`]);
```
MahboobehMohammadi: @shayanbinary why we have to change grid colors manually? can it be handled only using css?

morteza-binary: @MahboobehMohammadi because changing the CSS class affects the grid color by a delay.

PR #656, opened 2019-02-11
Morteza/multi_instance

morteza-binary:


src/components/ChartTypes.jsx
```diff
@@ -65,12 +66,13 @@ const ChartTypes = ({
...
-    Type: chartType.type,
-    setOpen: chartType.menu.setOpen,
-    onChange: chartType.setTypeFromUI,
+export default connect(({ chartType, state }) => ({
+    chartId            : state.chartId,
+    ChartTypeMenu      : chartType.ChartTypeMenu,
+    ChartTypeList      : chartType.ChartTypeList,
+    menuOpen           : chartType.menu.open,
+    onChange           : chartType.setTypeFromUI,
+    setOpen            : chartType.menu.setOpen,
```
MahboobehMohammadi: @morteza-binary nitpick- I think it's better to keep the code consistent with the rest of code.

morteza-binary: @MahboobehMohammadi You're right, but since all the projects of frontend should follow the same code style, IMO it'd be better we also follow the `binary2.0` and `binary-static` (which are the main project of the team) code style and make the changes step by steps in each PR.

PR #658, opened 2019-02-13
Mahboobeh/last_digits_stats_fixes

MahboobehMohammadi:


sass/components/_last-digits.scss
```diff
@@ -1,63 +1,72 @@
...
+        }
+        .cq-bar {
             width: 10px;
-            text-align: center;
+            @include themify($themes) {
+                background-color: themed('LastdigitsBarBg');
+            }
+            position: absolute;
+            bottom: 40px;
+            -webkit-transition: height 0.8s ease-in-out; /* For Safari 3.1 to 6.0 */
```
shayanbinary: @MahboobehMohammadi doesn't `POSTCSS` do this?

sass/components/_last-digits.scss
```diff
@@ -1,63 +1,72 @@
...
+            transition: height 0.8s ease-in-out;
+            border-bottom-left-radius: 16px;
+            border-bottom-right-radius: 16px;
+            border-top-left-radius: 16px;
+            border-top-right-radius: 16px;
```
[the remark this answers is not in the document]
MahboobehMohammadi: @shayanbinary i have no idea:))

src/components/LastDigitStats.jsx
```diff
@@ -3,19 +3,20 @@ import { connect } from '../store/Connect';
 import '../../sass/components/_last-digits.scss';
 
 const Bar = ({ x, bar }) => (
-    <div className={`cq-bar ${bar.cName}`} style={{ height: bar.height, left : (x * 35), '--bar-gradiant': `${bar.gradiantLine}%` }}>
+    <div className={`cq-bar ${bar.cName}`} style={{ height: `${bar.height * 5}%`, left : (x * 35) }}>
```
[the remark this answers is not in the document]
MahboobehMohammadi: @shayanbinary i dont know what you mean but they have to have absolute position. idn if it has conflict with flex-box and as you said they have dynamic hights

sass/components/_last-digits.scss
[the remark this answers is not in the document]
MahboobehMohammadi: @shayanbinary didnt know that.

src/components/LastDigitStats.jsx
morteza-binary: @MahboobehMohammadi Would you plz put each attribute in a separate line? it makes the code more readable.
Also, It's more readable to use variables for calculated values in then use the variables in the JSX tags.

src/components/LastDigitStats.jsx
```diff
@@ -3,19 +3,20 @@ import { connect } from '../store/Connect';
...
+        <div className={`cq-bar-value ${bar.cName}`} style={{ opacity: `${bar.height === 0 ? 0 : 1}` }}>{`${bar.height.toFixed(1)}%`}</div>
```
morteza-binary: @MahboobehMohammadi ditto

morteza-binary: @MahboobehMohammadi It'd be better to add a class to the `className` when `bar.height` is `0` instead of using style directly in the code?

src/store/LastDigitStatsStore.js
```diff
@@ -60,22 +67,33 @@ export default class LastDigitsStore {
...
+        }
     }
 
     @action.bound updateBars() {
         const min = Math.min(...this.digits);
         const max = Math.max(...this.digits);
         this.digits.forEach((digit, idx) => {
-            this.bars[idx].height = Math.round(((this.maxHeight - this.minHeight) * (digit - min) / (max - min)) + this.minHeight);
-            this.bars[idx].gradiantLine = this.bars[idx].height > this.gradiantLine ? ((this.bars[idx].height * this.gradiantLine) / this.maxHeight) : 0;
+            this.bars[idx].height = digit / 10;
```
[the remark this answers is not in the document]
MahboobehMohammadi: @morteza-binary what `count`? you mean use a variable?

src/components/LastDigitStats.jsx
MahboobehMohammadi: @morteza-binary we usually let eslint handle this kindda things. if eslint don't show errors means it's readable i think. 
i myself don't like separating  every props in single line. harder for me to read them.

morteza-binary: @MahboobehMohammadi Unfortunately there isn't the related rules enabled in our project so eslint doesn't manage it.

MahboobehMohammadi: @morteza-binary so we should add releated rules if there are there and have eslint to handle these changes, so code looks like the same everywhere.

morteza-binary: @MahboobehMohammadi Unfortunately, we can't add it the rule now, because it causes a lot of errors. I'm planning to add a card for refactoring the project style and rules in the further next sprints. But now we can follow it and instead of eslint.

PR #659, opened 2019-02-14
Shayan/historical_data_issues

shayan-binary:


src/components/Chart.jsx
```diff
@@ -98,7 +101,9 @@ class Chart extends Component {
                                         <CurrentSpot />
                                     </RenderInsideChart>
                                     <div className="cq-top-ui-widgets">
-                                        <TopWidgets />
+                                        {React.Children.map(<TopWidgets />, child => React.cloneElement(child, {
+                                            MainProp: {} }))}
```
shayanbinary: @morteza-binary I expect to pass `MainProp` here to it's children

PR #660, opened 2019-02-14
handle hiding icons by classname beside of disply

shayanbinary:


src/store/StudyLegendStore.js
```diff
@@ -52,7 +52,7 @@ export default class StudyLegendStore {
             if (panel) {
                 // Hide the up arrow from first indicator to prevent user
                 // from moving the indicator panel above the main chart
-                this.stx.panels[panel].up.style.display = 'none';
+                this.stx.panels[panel].up.className = 'stx-btn-panel';
```
shayanbinary: I don't know how, but something rapidly, change the style of this element, so I have to handle it by `class name`

MahboobehMohammadi: @shayanbinary seems its overriding `display` inside `drawPanels` of `chartiq line:36509`  so instead of changing its class which might change later using library you could append the function to `drawPanels` as well.

PR #661, opened 2019-02-14
Ashkan/control_menu_ui

ashkanx:


app/app.scss
```diff
@@ -1,5 +1,13 @@
 /* Used for app only. Do not deploy with library! */
 
+html {
+    font-size: 62.5%; /* 1rem = 10px */
```
morteza-binary: @ashkanx the `font-size` will be overwritten by `.smartcharts` class to `12px` therefore the values of `font-size` in each classes in `_chart-controls.scss` will be `12px` instead of `10px`. Also the values of left/right paddings will be `19.2px` instead of `16px`.

sass/_variables.scss
```diff
@@ -4,9 +4,9 @@
 
 $base-font: 'Roboto', helvetica, sans-serif !default;
 
-$font-size-base: 12px !default;
-$font-size-button: 11px !default;
-$font-size-toolbar: 14px !default;
+$font-size-base: 1em !default;
+$font-size-button: 1em !default;
+$font-size-toolbar: 1em !default;
```
morteza-binary: @ashkanx Should these variable be `1.2em`, `1.1em` and `1.4em` respectively? or all of them have been changed to `10px`in the new design?

ashkanx: I'm not sure yet. still waiting for UI team

PR #662, opened 2019-02-15
Fix symbol change issue

MahboobehMohammadi:


src/feed/Feed.js
```diff
@@ -148,7 +148,10 @@ class Feed {
             }
 
             subscription.onChartData((tickResponse) => {
-                this._appendChartData(tickResponse, key, comparisonChartSymbol);
+                // Append comming ticks to chart only if it belongs to selected symbol after symbol changes
+                if (isComparisonChart || symbol === this._stx.chart.symbol) {
```
morteza-binary: @MahboobehMohammadi I don't think it solves the issue whenever there are comparison charts on the main chart because the value of `isComparisonChart` is `symbol !== this._stx.chart.symbol` then this `if` statement is always `true`.

MahboobehMohammadi: @morteza-binary no. its different. whenever this function get called, it could being called from comparison stream or chart stream. this `isComparisonChart` is determine who is calling it. so its not what you think. its false whenever the tick is coming for symbol.

PR #663, opened 2019-02-15
Morteza/resizing

morteza-binary:
Maintain the scale of the chart and scroll after resizing


src/chartiq_injections/resizing.js
```diff
@@ -1,14 +1,14 @@
 export const maintainSpanSize = () => {
     CIQ.ChartEngine.prototype.prepend('resizeChart', function () {
-        if (this.chart.xaxis && this.chart.xaxis.length) {
+        if (this.isAutoScale && this.chart.xaxis && this.chart.xaxis.length) {
```
MahboobehMohammadi: @morteza-binary the better name is `AutoSize`, you dont need to add `is` to it.

src/components/ChartSetting.jsx
```diff
@@ -177,23 +188,25 @@ const ChartSetting = ({
...
+    setAutoScale       : s.setAutoScale,
+    setHistorical      : s.setHistorical,
+    setLanguage        : s.setLanguage,
+    setPosition        : s.setPosition,
+    setTheme           : s.setTheme,
+    setView            : s.setView,
+    showCountdown      : s.showCountdown,
+    theme              : s.theme,
+    view               : s.view,
 }))(ChartSetting);
```
MahboobehMohammadi: @morteza-binary could you pls do refactoring and adding new codes in separate comments , it's hard to review cause its hard to find new codes among last refactored codes.

src/chartiq_injections/resizing.js
morteza-binary: @MahboobehMohammadi It'd be better the boolean variables have `is`, `should` or `has` in their names. it helps developers to know the type of the variable.
It's one of the naming convention which we should follow in the future refactorying.

src/components/ChartSetting.jsx
```diff
@@ -105,6 +107,15 @@ const ChartSetting = ({
...
                     </div>
+                    <div className="ciq-list-item">
+                        <span className="ciq-icon-text">{t.translate('Auto Scale')}</span>
+                        <div className="ciq-action">
+                            <Switch
+                                value={isAutoScale}
+                                onChange={setAutoScale}
+                            />
+                        </div>
+                    </div>
```
MahboobehMohammadi: @morteza-binary i think its better to sort settings alphabetically since its getting big. cause now they has no meaningful order.

src/chartiq_injections/resizing.js
MahboobehMohammadi: @morteza-binary `auto something` true or false will context the meaning but its a nitpick anyway.

src/components/ChartSetting.jsx
morteza-binary: @MahboobehMohammadi Actually it'd be better to apply the style changes in ongoing PRs instead of postponing everything to one day in the future.

src/store/ChartSettingStore.js
```diff
@@ -177,4 +180,13 @@ export default class ChartSettingStore {
         }, 10);
         this.menu.setOpen(false);
     }
+
+    @action.bound setAutoScale(value) {
+        if (this.isAutoScale === value) { return; }
+
+        this.isAutoScale = value;
+        logEvent(LogCategories.ChartControl, LogActions.ChartSetting, `${value ? 'Show'  : 'Hide'} AutoScale`);
```
MahboobehMohammadi: @morteza-binary here i think it's better if you log something like `Change AutoScale to ${AutoScale}`.

src/components/ChartSetting.jsx
MahboobehMohammadi: @morteza-binary yes same PRs, but should be in different comments cause otherwise reviewing will be impossible.

src/chartiq_injections/resizing.js
morteza-binary: @MahboobehMohammadi it'd be better to follow the company naming conventions instead. BTW I don't think it is grammatically wrong.

PR #664, opened 2019-02-16
Shayan/add_marker_between_intervals

shayanbinary:


app/test.jsx
```diff
@@ -115,26 +125,37 @@ class App extends React.Component {
             high,
             low,
         }] : [];
+
+        const MarkerTimes = [];
+
+        for (let i = 0; i < 5; i++) {
+            MarkerTimes.push(moment().utc().second(0).subtract(i + 3, 'minutes')
```
shayanbinary: I add sample marker here, as this file is a test file I don't think that makes any problem

app/test.jsx
```diff
@@ -115,26 +125,37 @@ class App extends React.Component {
...
-                    />
+                        granularity={this.state.granularity}
+                    >
+                        {MarkerTimes.map(x => (
+                            <Marker
+                                key={x}
+                                className="chart-marker-line"
+                                x={x}
+                                xPositioner="epoch"
+                                yPositioner="top"
```
morteza-binary: @shayanbinary IMO, we can handle the `yPositioner` to `top` also in `MarkerStore` after adding the invisible bar, because the third-party applications don't care about it that the `x` exists exactly in the `x-axis` or not.

PR #670, opened 2019-02-26
Mahboobeh/stx is undefined error fix

MahboobehMohammadi:


src/store/LastDigitStatsStore.js
```diff
@@ -51,7 +51,7 @@ export default class LastDigitStatsStore {
 
             this.updateChartMargin(150);
 
-            if (this.stx.masterData && this.stx.masterData.length >= this.count) {
+            if (this.context && this.stx.masterData && this.stx.masterData.length >= this.count) {
```
morteza-binary: @MahboobehMohammadi would it be better to wait for `context` like other dependent stores then initializing the store and component?

MahboobehMohammadi: @morteza-binary this component is being initialized from outside. when this method called the context is initialized, since i didnt know what caused the error reported i check for context not to be null here though im pretty sure this never happens.

morteza-binary: @MahboobehMohammadi it happens when the value of `showLastDigitStats` is true when the chart is loading. Also, it is possible to happen when the page is refreshing and the value of `showLastDigitStats` is true. 
Anyway, it has been handled completely in `BottomWidgetsContianer` component and prevent to show widget until the `context` gets ready.

MahboobehMohammadi: @morteza-binary thanks. i guess you can close this PR though

morteza-binary:
Handled in #672

PR #671, opened 2019-02-26
Mahboobeh/ Add import and export layout functionality to smartchart

MahboobehMohammadi:


src/store/ChartState.js
```diff
@@ -20,6 +24,9 @@ class ChartState {
         this.mainStore = mainStore;
         this.chartStore = mainStore.chart;
         when(() => this.context, this.onContextReady);
+        reaction(() => this.onExportLayout, this.ExportLayout.bind(this));
```
morteza-binary: @MahboobehMohammadi You can call `ExportLayout`, `ImportLayout` and `CleanChart` in `updateProps` function. It will be invoked whenever one of the props is changed.

src/store/ChartState.js
```diff
@@ -11,6 +11,10 @@ class ChartState {
     @observable isConnectionOpened;
     @observable settings;
     @observable showLastDigitStats;
+    @observable onExportLayout;
+    @observable clearChart;
+    @observable importLayout;
```
morteza-binary: @MahboobehMohammadi it'd be better to name it `importedLayout`

src/store/ChartState.js
```diff
@@ -152,6 +171,60 @@ class ChartState {
             this.stxx.draw();
         }
     }
+
+    ClearChart() {
```
morteza-binary: @MahboobehMohammadi Should it be `clearChart`? it's a method of a class.

src/store/ChartState.js
```diff
@@ -152,6 +171,60 @@ class ChartState {
...
+        // Remove indiactors
+        for (const id in this.stxx.layout.studies) {
+            const sd = this.stxx.layout.studies[id];
+            CIQ.Studies.removeStudy(this.stxx, sd);
+        }
+        this.stxx.clearDrawings();
+        this.mainStore.crosshair.setCrosshairState(0);
+
+        this.mainStore.chart.changeSymbol(this.stxx.chart.symbol, 0);
+        this.mainStore.chartType.setType('mountain');
```
morteza-binary: It'd be better to create a constants file for all default value in the project. It should be done probably on another card. Please if it's possible for you, add a TODO before the line to help us whenever we want to refactor the project.

src/store/ChartState.js
```diff
@@ -152,6 +171,60 @@ class ChartState {
...
+    }
+
+    ImportLayout() {
```
morteza-binary: @MahboobehMohammadi Also, it should be `importLayout`.

src/store/ChartState.js
```diff
@@ -152,6 +171,60 @@ class ChartState {
...
+        if (!this.importLayout) return;
+        this.stxx.importLayout(this.importLayout, {
```
morteza-binary: @MahboobehMohammadi What happens if the developer passes an empty object by mistake?

src/store/ChartState.js
```diff
@@ -152,6 +171,60 @@ class ChartState {
...
+                }, 500);
+            },
+        });
+
+        this.mainStore.crosshair.setCrosshairState(this.importLayout.crosshair);
+        this.mainStore.chart.changeSymbol(this.stxx.chart.symbol, this.importLayout.interval * 60);
+        this.mainStore.chartType.setType(this.importLayout.chartType);
+    }
+
+    ExportLayout() {
```
morteza-binary: @MahboobehMohammadi `exportLayout`?

src/store/ChartState.js
MahboobehMohammadi: @morteza-binary i dont think developer with send an empty obj! it's either null or the obj that he exported!!

morteza-binary: @MahboobehMohammadi Anyway, I'd be better to handle race conditions.

PR #672, opened 2019-02-26
Morteza/last_digit_indicator

morteza-binary:


README.md
```diff
@@ -105,6 +105,7 @@ isConnectionOpened | Sets the connection status. If set, upon reconnection smart
 onMessage | SmartCharts will send notifications via this callback, should it be provided. Each notification will have the following structure: `{ text, type, category }`.
 isAnimationEnabled | Determine whether chart animation is enabled or disabled. It may needs to be disabled for better performance. Defaults to `true`.
 showLastDigitStats | Shows last digits stats. Defaults to `false`.
+bottomWidgets | Render function for bottom widgets. Refer to [Customising Components](#customising-components).
```
MahboobehMohammadi: @morteza-binary could you please add it beside `topwidget` and also provide an example like we did for `topwidget`

src/components/BottomWidgetsContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+                className="cq-bottom-ui-widgets"
+                style={styles}
+            >
+                { children }
+            </div>
+        );
+    }
+}
+
+BottomWidgetsContainer.propTypes = {
```
MahboobehMohammadi: @morteza-binary we are not using prop-type? why you need it here?

src/components/BottomWidgetsContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes   from 'prop-types';
+import { connect } from '../store/Connect';
+
+import '../../sass/components/_bottom-widget-container.scss';
+
+
+class BottomWidgetsContainer extends React.Component {
+    componentDidUpdate() {
+        if (React.Children.count(this.props.children)) {
+            this.props.updateChartMargin(200);
```
MahboobehMohammadi: @morteza-binary doesn't seems right to use a fix number here. i used it in last digit cause i knew the height of last digit component but what if user add another component?

src/components/BottomWidgetsContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+            return null;
+        }
+
+        const styles = {
+            top,
+            bottom,
+        };
+
+        return (
+            <div
```
MahboobehMohammadi: @morteza-binary i belive you could do st like `isReadyToShow && ....` here.

MahboobehMohammadi: @morteza-binary i believe you could do something like `isReadyToShow && ....` here.

src/components/BottomWidgetsContainer.jsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+BottomWidgetsContainer.propTypes = {
+    bottom           : PropTypes.number,
+    isReadyToShow    : PropTypes.bool,
+    top              : PropTypes.number,
+    updateChartMargin: PropTypes.func.isRequired,
+};
+
+BottomWidgetsContainer.defaultProps = {
```
MahboobehMohammadi: @MahboobehMohammadi i think you should follow smart chart code style. its so confusing for next user to read different codes

src/store/index.js
```diff
@@ -46,4 +47,5 @@ export default class MainStore {
     drawingCursor = new DrawingCursorStore(this);
     chartTable = new ChartTableStore(this);
     lastDigitStats= new LastDigitStatsStore(this);
+    bottomWidgetsContainerStore = new BottomWidgetsContainerStore(this);
```
MahboobehMohammadi: @morteza-binary you can remove the store from end of the object name here. pls just use `bottomWidgetsContainer`

src/components/BottomWidgetsContainer.jsx
morteza-binary: @MahboobehMohammadi It should be added to all the project to make the code more clear.

src/components/BottomWidgetsContainer.jsx
```diff
@@ -0,0 +1,63 @@
+import React       from 'react';
+import PropTypes   from 'prop-types';
+import { connect } from '../store/Connect';
+
+import '../../sass/components/_bottom-widget-container.scss';
+
+
+class BottomWidgetsContainer extends React.Component {
```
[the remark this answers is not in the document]
morteza-binary: @MahboobehMohammadi In this way, we keep the code more clear and organized.

src/components/BottomWidgetsContainer.jsx
morteza-binary: @MahboobehMohammadi IMO, it's better to keep the height of the bottom widget fix. 
@ashkanx What's your idea? 

BTW, If we want to calculate the wide of the children and use it here, we have to use `ref` for each element in the children to get its height that IMO, it is not a good idea.

src/components/BottomWidgetsContainer.jsx
morteza-binary: @MahboobehMohammadi yes, it's possible but in this way, it doesn't need any curly brackets in return.

src/components/BottomWidgetsContainer.jsx
morteza-binary: @MahboobehMohammadi regarding the discussion with Ashkan, we're planning to refactor the code style in this project in the next sprint so it's better to do them for the new component.

src/components/BottomWidgetsContainer.jsx
ashkanx: I think it can't be a fixed value eventually. it should be relative to screen size and other active components in the screen. but probably we should cover that in another card including UI team as well.

MahboobehMohammadi: @morteza-binary why just not use a Height prop so anybody use it just set the height. It’s not like it contains dynamic children.

PR #673, opened 2019-02-28
Shayan/fix scroll on add favorite item

shayanbinary:


src/store/CategoricalDisplayStore.js
```diff
@@ -160,13 +158,15 @@ export default class CategoricalDisplayStore {
     }
 
     @computed get favoritesCategory()  {
+        this.pauseScrollSpy = true;
```
morteza-binary: @shayanbinary  Changing `this.pauseScrollSpy` here doesn't fix the issue.

shayanbinary: @morteza-binary I increase the delay, seems it works fine

PR #678, opened 2019-03-06
Mahboobeh/Fix countdown lable position

MahboobehMohammadi:


src/chartiq_injections/currentHR.js
```diff
@@ -1,28 +1,83 @@
+/* eslint-disable */
 export const drawCurrentPriceLine = () => {
-    CIQ.ChartEngine.prototype.prepend('drawCurrentHR', function () {
-        const mainSeriesRenderer = this.mainSeriesRenderer || {};
-        if (mainSeriesRenderer.noCurrentHR) return;
-        for (const chartName in this.charts) {
-            const chart = this.charts[chartName];
-            const panel = chart.panel;
+CIQ.ChartEngine.prototype.drawCurrentHR = function () {
```
morteza-binary: @MahboobehMohammadi According to the ChartIQ documentation is not a good idea to overwrite the main functionality of the chart engine, it'd be better to use injections. is there any obstacle that makes you decide to overwrite this function?

MahboobehMohammadi: @morteza-binary there is no other way to do it. the position of current price label is being calculated here in a twisted way! the only place we can get its position is here. it can not be done using injections. and the label position is not consistent, base of having comparison or scroll chart or other thing.

morteza-binary: @MahboobehMohammadi I see, but have you tried `stx.pixelFromPrice(stx.currentQuote().Close)`? as I checked, its result is always as same as the value of the variable you're adding to the chart. (FYI, I've checked line and candle chart types)

MahboobehMohammadi: @morteza-binary as you can see in my first commits yes i did but when we have comparsions then this wont work cause everything change to percent and.... .

morteza-binary: @MahboobehMohammadi I see, but what about `stx. pixelFromTransformedValue` ? also you can access the transformed value of the current quote by `stx.currentQuote().transform.Close`.

MahboobehMohammadi: we can not use `pixelFromTransformedValue` or `pixelFromPrice` because they show the position of current close so when you scroll back it is the wrong value.

PR #685, opened 2019-03-15
add marker after data appending data

shayanbinary:


shayanbinary:
This modification already applied on another PR

PR #691, opened 2019-03-25
Shayan/Make sure markers stay on top of other elements including barrier

shayanbinary:
we try to set the same parent for chart and markers so that we be able to handle it by `z-index`, for the test link, all markers set `z-index: 36;` so they stay on the top


src/components/Chart.jsx
```diff
@@ -93,6 +93,7 @@ class Chart extends Component {
                             <div className="ciq-chart-area">
                                 <div className="ciq-chart">
                                     <RenderInsideChart at="holder">
+                                        {children}
```
morteza-binary: @shayanbinary Regarding the description in the chartIQ site, it'd be better to place the markers in the `subholder`.
image

shayanbinary:
at last, we come to this conclusion that moves the `barrier` to the `<RenderInsideChart at="subholder">` element

app/test.jsx
```diff
@@ -406,8 +420,13 @@ class App extends Component {
                 </div>
                 <div className="action-section">
                     <div className="form-row">
-                        Line Markers <br />
-                        <button type="button" onClick={this.onCircleMarker}>Toggle</button>
+                        Markers <br />
+                        <select onChange={this.onAddMArker}>
```
shayanbinary: Change this for testing purpose

src/store/BarrierStore.js
```diff
@@ -238,11 +240,14 @@ export default class BarrierStore {
     }
 
     _calcBottomShade(barrier) {
-        return this.chart.panel.height - barrier.top;
+        const studyPanel = this.mainStore.chart.modalNode.querySelector('.stx-panel-study');
+
+        return this.chart.panel.height - barrier.top - (!studyPanel ? LINE_OFFSET_HEIGHT : 0);
```
shayanbinary: @morteza-binary actually resizing not working, I found this solution.

src/store/BarrierStore.js
```diff
@@ -1,11 +1,13 @@
-import { observable, action, computed } from 'mobx';
+import { observable, action, computed, when } from 'mobx';
 import PriceLineStore from './PriceLineStore';
 import ShadeStore from './ShadeStore';
 import PendingPromise from '../utils/PendingPromise';
 import PriceLine from '../components/PriceLine.jsx';
 import Shade from '../components/Shade.jsx';
 import { isValidProp } from '../utils';
 
+const LINE_OFFSET_HEIGHT = 30;
```
morteza-binary: Why is it a global variable?

src/store/BarrierStore.js
```diff
@@ -60,6 +64,23 @@ export default class BarrierStore {
 
         this.HighPriceLine = this._high_barrier.connect(PriceLine);
         this.LowPriceLine = this._low_barrier.connect(PriceLine);
+
+        when(() => this.stxx, this.onContextReady);
+    }
+
+    get context() { return this.mainStore.chart.context; }
```
morteza-binary: @shayanbinary It'd be better to move the get properties before the constructor.

src/store/BarrierStore.js
```diff
@@ -128,16 +150,19 @@ export default class BarrierStore {
...
 
             return newPrice;
         };
     }
 
-    get context() { return this.mainStore.chart.context; }
-    get stx() { return this.context.stx; }
-    get chart() { return this.stx.chart; }
+    @action.bound calculateLabelRight() {
+        this.labelRight = this.mainStore.chart.stxx.chart.yAxis.width * -1;
```
morteza-binary:
```suggestion
        this.labelRight = this. yAxisWidth * -1;
```

src/store/BarrierStore.js
shayanbinary: @morteza-binary the shade jump size pixel is fixed. and it's around `30px`

morteza-binary: @ I see, but it can be a constant property of the class instead of being global unless it is used by other classes too.

shayanbinary: @morteza-binary aha, I get your point

PR #692, opened 2019-03-25
Mahboobeh/ update props on app side when they're being change by layout import and e…

MahboobehMohammadi:
…xport


src/store/TimeperiodStore.js
```diff
@@ -19,6 +19,8 @@ export default class TimeperiodStore {
     get isTick() { return this.timeUnit === 'tick'; }
     @observable timeUnit = null;
     @observable interval = null;
+    @observable onGranularityChange;
```
morteza-binary: it doesn't need to be observable.

src/store/ChartTypeStore.js
```diff
@@ -247,4 +246,5 @@ export default class ChartTypeStore {
     }
 
     @observable type = getChartTypes()[0];
+    @observable ocChartTypeChanged;
```
morteza-binary: I doesn't need to be observable.

src/components/ChartTypes.jsx
```diff
@@ -75,4 +78,5 @@ export default connect(({ chartType, state }) => ({
     setOpen            : chartType.menu.setOpen,
     showAggregateDialog: chartType.showAggregateDialog,
     Type               : chartType.type,
+    onChartTypeChanged : chartType.onChange,
```
morteza-binary: @MahboobehMohammadi it'd be better to choose another name to covey that it's adding the passed`onChange` function from parent to the `ChartTypeStore`.

src/components/Timeperiod.jsx
```diff
@@ -134,4 +138,5 @@ export default connect(({ timeperiod: s, state }) => ({
     setOpen         : s.menu.setOpen,
     TimePeriodMenu  : s.TimePeriodMenu,
     timeUnit_display: s.timeUnit_display,
+    onGranularityChange: s.onChange,
```
morteza-binary: @MahboobehMohammadi it'd be better to choose another name to covey that it's adding the passed `onChange` function from parent to the `TimeperiodStore`.

PR #699, opened 2019-04-01
Mahboobeh/Import layout after chart is loaded

MahboobehMohammadi:


chartiq/chartiq.js
```diff
@@ -21,7 +21,7 @@
 
 (function() {
 	var trialExpiration = undefined;
-	var licenseExpiration = "2019/04/01";
+	var licenseExpiration = "2020/04/01";
```
morteza-binary: @MahboobehMohammadi Should we get the updated version instead of updating it manually?

MahboobehMohammadi: @morteza-binary i dont know when we get the new updates, but yes i think we shouldnt change manulay though seems it has no effects but print a console error.
by the way i think we shouldnt update now. since we had a lot of changes lets first fix our errors and get stable then we can update if there is a new one, cause we usually face some issues after updating too.

morteza-binary: @MahboobehMohammadi 
I think it should be possible to just renew the license without upgrading the library version.

morteza-binary: @MahboobehMohammadi Would you please to revert the license expiry date to the previous one to merge the PR?

ashkanx: I sent an email to their support team

PR #701, opened 2019-04-02
Mahboobeh/Hide chart control widget when its set to null

MahboobehMohammadi:


src/components/Chart.jsx
```diff
@@ -129,7 +129,7 @@ class Chart extends Component {
                                         }
                                     </BottomWidgetsContainer>
                                 </div>
-                                <ChartControls widgets={chartControlsWidgets} />
+                                { chartControlsWidgets !== null && <ChartControls widgets={chartControlsWidgets} /> }
```
morteza-binary:
```suggestion
                                { chartControlsWidgets !== null && 
                                    <ChartControls widgets={chartControlsWidgets} /> 
                                 }
```

src/store/ChartState.js
```diff
@@ -95,6 +96,11 @@ class ChartState {
                 }
             }
         }
+
+        if (chartControlsWidgets !== this.chartControlsWidgets) {
+            this.chartControlsWidgets = chartControlsWidgets;
+            if (this.stxx) this.mainStore.chart.updateHeight();
```
morteza-binary: @MahboobehMohammadi It's possible to invoke the `updateHeight` function in `Chart.jsx` when the component is updated and the value of `chartcontrolWidgets` is null. It doesn't need to add an unusable observable property to `ChartState`.

Note, if you want to change the code to the above solution, you should pass the control component or a true/false flag to the `updateHeight` function to find out that the control widget exists or not.

MahboobehMohammadi: @morteza-binary yes it doesnt need to be observable but still i will keep the previous value of widget so i wont call `updateHeight` when any prop change and using flags is not a very clear so i usually avoid them.

morteza-binary: @MahboobehMohammadi You can check the previous and current value of `chartControlWidgets` in `componentDidUpdate` lifestyle and then decide to invoke `updateHeight` or not.

MahboobehMohammadi: @morteza-binary yes but we are handling everything else with prop. why break the pattern here?! its just different approach, i think its fine

PR #711, opened 2019-04-05
Morteza/flying_markers

morteza-binary:
Fixed the flying markers issue


src/store/MarkerStore.js
```diff
@@ -112,7 +112,7 @@ export default class MarkerStore {
                 );
                 this.stxx.createDataSet();
 
-                if (this.yPositioner !== 'top' && this.yPositioner !== 'on_candle') {
+                if (this.yPositioner !== 'value' && this.yPositioner !== 'on_candle') {
```
shayanbinary: @morteza-binary, of course, seems this is the solution, but I just wonder why we decide to contain `top` and `on_candle` on this condition

morteza-binary: @shayanbinary Definitely, the developer of that part of code can answer your question :-P

PR #717, opened 2019-04-09
Mahboobeh/ Fix chart issues when used by contracts

MahboobehMohammadi:


src/store/TimeperiodStore.js
```diff
@@ -113,7 +113,9 @@ export default class TimeperiodStore {
     }
 
     @action.bound updateProps(onchange) {
-        this.onGranularityChange = onchange;
+        if (this.mainStore.state.granularity !== undefined) {
```
morteza-binary:
```suggestion
        if (onChange !== undefined) {
```

PR #718, opened 2019-04-10
sentence case

ashkanx:


src/components/Views.jsx
```diff
@@ -88,7 +88,7 @@ const Views = ({
                                     />
                                 </span>
                             )
-                            : <span className="add-new" onClick={main}> {t.translate('Add New')} </span>
+                            : <span className="add-new" onClick={main}> {t.translate('Add new')} </span>
```
morteza-binary: @ashkanx Would you please update translations file via `yarn translations`?

PR #726, opened 2019-04-18
Countdown for time until market opens

nazaninreihani:


src/components/Timer.jsx
```diff
@@ -0,0 +1,33 @@
...
+            () => this.tick(),
+            1000,
+        );
+    }
+    componentWillUnmount() {
+        clearInterval(this.timerInterval);
+    }
+    tick() {
+        this.setState({
+            time: new Date().getTime(),
```
morteza-binary: @nazaninreihani should it be server-time? what happens if the client's machine datatime is not accurate?

src/components/Timer.jsx
```diff
@@ -0,0 +1,33 @@
...
+        });
+    }
+    // 86400000 = 24 hour * 60 min * 60s * 1000ms
+    render() {
+        return (
+            <span>
+                {this.props.symbolOpenTime.symbolOpenTime ? displayMilliseconds(86400000 - (this.state.time - this.props.symbolOpenTime.symbolOpenTime)) : '--'}
```
morteza-binary: @nazaninreihani this algorithm just works when the open time is in 24 hours. 
@ashkanx Do we have any cases that the open time is after the next 24 hours?

morteza-binary:
@nazaninreihani would you please update translations files too?

src/store/ChartTitleStore.js
```diff
@@ -51,7 +54,17 @@ export default class ChartTitleStore {
     @computed get decimalPlaces() { return this.mainStore.chart.currentActiveSymbol.decimal_places; }
     @computed get isShowChartPrice() { return this.mainStore.chart.isChartAvailable; }
     @computed get tradingTimes() { return this.mainStore.chart.tradingTimes; }
-    @computed get symbolOpenTime() { return this.tradingTimes._tradingTimesMap[this.currentSymbol.symbol].symbolOpenTime; }
+    @computed get symbolOpenTime() {
+        const times = this.tradingTimes._tradingTimesMap[this.currentSymbol.symbol].times;
+        const now = this.serverTime.getLocalDate().getTime();
+        let openTime = times ? times.find(time => time.open.getTime() > now) : null;
+
+        if (!isObjectOfTypeDate(openTime)) {
```
morteza-binary:
```suggestion
        if (!(openTime instanceof Date)) {
```

nazaninreihani: I thought this is going to filter invalid dates unlike instance of date, but I tested it now and it didn't do that :woman_facepalming:

PR #731, opened 2019-04-21
Shayan/Add zoom in/out widget

shayan-binary:


app/test.jsx
```diff
@@ -405,6 +414,10 @@ class App extends Component {
                     </SmartChart>
                 </div>
                 <div className="action-section">
+                    <div className="form-row">
+                        Widget <br />
```
ahkuma: Instead of breaking, I think you should use a `<p>` tag instead, looks cleaner that way.

sass/_themes.scss
```diff
@@ -218,7 +218,9 @@ $themes: (
         ChartHistoryPickerBodyActiveBg: $COLOR_DARK_BLUE_5,
         ChartHistoryPickerBodyActiveColor: $COLOR_WHITE,
         ChartHistoryPickerBodyActiveBorder: $COLOR_DARK_BLUE_5,
-        PaginationLoaderStripeColor: $COLOR_LIGHT_BLACK_4_SOLID
+        PaginationLoaderStripeColor: $COLOR_LIGHT_BLACK_4_SOLID,
+        WidgetBg: $COLOR_LIGHT_BLACK_4_SOLID,
```
ahkuma: A bit nit picky but I think `WidgetBackground` would be a bit more readable or at least `WidgetBG`

app/test.jsx
```diff
@@ -331,6 +331,14 @@ class App extends Component {
         this.setState({ markers });
     }
 
+    onWidget = () => {
+        const { widget } = this.state;
+
+        this.setState({
+            widget: !widget,
```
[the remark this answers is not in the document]
shayanbinary: yeah of course, but this return an Eslint issue, which I had to way, disable the Eslint or fix it that way

app/test.jsx
shayanbinary: this is our `test` file, so mostly we try to handle it very simple

sass/_themes.scss
shayanbinary: yep, as other word goes this way, I just try to follow the rules

src/components/Widget.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { connect } from '../store/Connect';
+import '../../sass/components/_ciq-widget.scss';
+
+import { ZoomInIcon, ZoomOutIcon, HomeIcon, ScaleIcon } from './Icons.jsx';
+
+const Widget = ({
```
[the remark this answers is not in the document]
shayanbinary: @morteza-binary any suggestion? :)

src/store/WidgetStore.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    onContextReady = () => { this.stx = this.mainStore.chart.context.stx; };
+
+    @action.bound onHome() {
+        this.stx.home();
+    }
+
+    @action.bound onScale() {
+        if (this.stateStore.scrollToEpoch) {
```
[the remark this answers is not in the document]
shayanbinary: so how about the current behavior of `scrollChartToLeft` in the `StateStore` when `scrollToEpoch` not set ? (`} else {this.stxx.chart.lockScroll = false;this.stxx.home();this.stxx.draw();}`)

sass/_themes.scss
```diff
@@ -431,6 +435,10 @@ $themes: (
     darkchartgrid: map-get(map-get($themes, 'dark'), 'ChartGrid');
     lightchartmountainborder: map-get(map-get($themes, 'light'), 'ChartMountainBorder');
     darkchartmountainborder: map-get(map-get($themes, 'dark'), 'ChartMountainBorder');
+    lightchartmountainbg: map-get(map-get($themes, 'light'), 'ChartMountainBg');
```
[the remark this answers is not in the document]
shayanbinary: yes of course, but it through an eslint error. if you don't mind, I can ignore this error.

src/store/ChartState.js
```diff
@@ -319,15 +323,16 @@ class ChartState {
         }
     }
 
-    scrollChartToLeft = () => {
+    scrollChartToLeft(epoch) {
```
morteza-binary: @shayanbinary the function should be defined as an arrow function because it's used as a callback function on [line 124](https://github.com/binary-com/SmartCharts/pull/731/files#diff-01fca571a632d65a132cb788015d68f1R124)

src/store/ChartState.js
```diff
@@ -379,10 +385,12 @@ class ChartState {
                 );
                 this.stxx.createDataSet();
             }
-            this.stxx.chart.lockScroll = true;
+
+            this.stxx.chart.lockScroll = false;
```
morteza-binary: @shayanbinary it should be `true`

src/store/NavigationWidgetStore.js
```diff
@@ -0,0 +1,42 @@
...
+    get stateStore() { return this.mainStore.state; }
+    get stxx() { return this.chart.stxx; }
+
+    constructor(mainStore) {
+        this.mainStore = mainStore;
+        when(() => this.mainStore.chart.context, this.onContextReady);
+    }
+
+    onContextReady = () => {
+        this.stx = this.mainStore.chart.context.stx;
```
morteza-binary: @shayanbinary what's the difference between  `context.stx` and `chart.stxx`? both of them are the same.

src/store/NavigationWidgetStore.js
```diff
@@ -0,0 +1,42 @@
+import { observable, action, when } from 'mobx';
+import { getUTCEpoch } from '../utils';
+
+export default class NavigationWidgetStore {
+    @observable stx;
+    @observable enableHome = false;
```
morteza-binary:
```suggestion
    @observable isHomeEnabled = false;
```

src/store/NavigationWidgetStore.js
```diff
@@ -0,0 +1,42 @@
...
+    get stateStore() { return this.mainStore.state; }
+    get stxx() { return this.chart.stxx; }
+
+    constructor(mainStore) {
+        this.mainStore = mainStore;
+        when(() => this.mainStore.chart.context, this.onContextReady);
+    }
+
+    onContextReady = () => {
+        this.stx = this.mainStore.chart.stxx;
```
morteza-binary: @shayanbinary we don't need this line anymore.

src/store/ChartState.js
```diff
@@ -380,13 +385,15 @@ class ChartState {
...
             this.stxx.maxMasterDataSize = 0;
+
             const tick = this.stxx.tickFromDate(startEntry.DT);
             const scrollAnimator = new CIQ.EaseMachine(Math.easeOutCubic, 1000);
             const scrollToTarget = this.stxx.chart.dataSegment.length;
             scrollAnimator.run((bar) => {
                 bar = Math.ceil(bar); // round-up for precision
                 const scroll = this.stxx.chart.dataSegment.length - bar;
                 if (scroll <= 2 || bar === scrollToTarget) {
+                    // console.log('A1');
```
morteza-binary: @shayanbinary Remove this line.

src/components/ui/Animation.js
```diff
@@ -85,7 +85,7 @@ export default function animateChart(stx, animationParameters, easeMachine) {
         chart = chart || this.chart;
         if (chart.lockScroll) {
             const lastAppendQuote     = appendQuotes[appendQuotes.length - 1];
-            const lastDataSegmentItem = chart.dataSegment[chart.dataSegment.length - 1];
+            const lastDataSegmentItem = chart.dataSegment ? chart.dataSegment[chart.dataSegment.length - 1] : null;
```
shayanbinary: @morteza-binary I am not sure about this, why this happen sometimes

morteza-binary: @shayanbinary because it's the chart is reinitialized and there isn't any data on `dataSegment`

src/store/ChartState.js
```diff
@@ -358,15 +362,17 @@ class ChartState {
         }
     }
 
-    scrollChartToLeft = () => {
+    scrollChartToLeft = (leftTick, force) => {
```
morteza-binary: @shayanbinary We don't need to add an extra parameter to the function since the function is invoked with `force` as `true` when `leftTick` is also passed. so you can't check whenever the passed parameter (`leftTick`) has value, the animation shouldn't be run. I have added a suggestion on `if` statement line.

src/store/ChartState.js
```diff
@@ -381,28 +387,45 @@ class ChartState {
...
-                     * the scroll value is off-screen or if the animator has reached target.
-                     * We check scroll <= '2' because sometimes the chart is scrolled so that the first
-                     * bar is partially hidden off-screen
-                     */
-                    scrollAnimator.stop();
-                    this.stxx.chart.entryTick = this.stxx.tickFromDate(startEntry.DT);
-                    this.stxx.chart.lockScroll = true;
-                    this.stxx.chart.isScrollLocationChanged = true; // set to true to draw markers
+
+            if (!force && !this.startEpoch) {
```
```suggestion
            if (!leftTick && !this.startEpoch) {
```

src/store/ChartState.js
```diff
@@ -362,9 +362,8 @@ class ChartState {
         }
     }
 
-    scrollChartToLeft = (date) => {
-        const scrollToEpoch = this.scrollToEpoch
-                            || (date && getUTCEpoch(date.DT));
+    scrollChartToLeft = (leftTick) => {
+        const scrollToEpoch = this.scrollToEpoch || getUTCEpoch(leftTick && leftTick.DT);
```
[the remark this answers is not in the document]
shayanbinary: oops, you are right

src/store/ChartState.js
shayanbinary: @morteza-binary navigation widget is not the only component use `scrollChartToLeft`, Feed.js use it as well. mean when you click on `purchase`, `scrollChartToLeft` called with a `leftTick` which is should run ScrollAnimator, that's why I add another prop

src/store/ChartState.js
shayanbinary: @morteza-binary regarding my previous comment, I put this

src/store/ChartState.js
morteza-binary: @shayanbinary Would you please let me know which line of `Feed.js` is using `scrollChartToLeft`?

app/test.jsx
```diff
@@ -356,6 +355,14 @@ class App extends Component {
         this.setState({ markers });
     }
 
+    onWidget = () => {
+        const { enabledNavigationWidget } = this.state;
+
+        this.setState({
```
nazanin-binary: Don't you think it's better to change it to the following?
`this.setState(prevState => ({ enabledNavigationWidget: !prevState.enabledNavigationWidget }))`

shayan-binary: @nazanin-binary nice shot, thanks.

morteza-binary:
Shayan, Would you please resolve the conflict?

PR #735, opened 2019-04-24
Fix issues when showing chart for contract result 

MahboobehMohammadi:


src/store/MarkerStore.js
```diff
@@ -201,7 +201,7 @@ export default class MarkerStore {
 
         this.stx.setMarkerTick(dummyMarker);
         this.tick = dummyMarker.tick;
-        if (dummyMarker.params.future) {
+        if (dummyMarker.params.future && this.mainStore.chart.xaxis && this.mainStore.chart.xaxis.length > 0) {
```
morteza-binary: @MahboobehMohammadi Would you please change the `if` statement to something like below to avoid adding `null` item to the masterData when the marker is in future?

```js
        if (dummyMarker.params.future) {
            this.isDistantFuture = true;

            if (this.mainStore.chart.xaxis && this.mainStore.chart.xaxis.length) {
                this.stx.futureTickIfDisplayed(dummyMarker);
                this.tick = dummyMarker.tick;
                if (this.tick !== null) {
                    this.isDistantFuture = false;
                }
            } else {
                this.hideMarker();
                return;
            }
        }
````

PR #736, opened 2019-04-24
Status of chart for when it's ready to be used by app

nazaninreihani:
by passing the prop getChartStatus to smartCharts (in the app using it) we can listen to the changes of the state of chart, status will be 'ready' when chart is loaded


src/store/ChartState.js
```diff
@@ -40,10 +41,12 @@ class ChartState {
         this.chartStore.feed.onPagination(this.setOnPagination.bind(this));
     };
 
-    @action.bound updateProps({ id, settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch, onExportLayout, clearChart, importedLayout, removeAllComparisons, isAnimationEnabled = true, showLastDigitStats = false, scrollToEpoch, scrollToEpochOffset = 0, zoom, chartControlsWidgets }) {
+    @action.bound updateProps({ id, settings, isConnectionOpened, chartStatusListener, symbol, granularity, chartType, startEpoch, endEpoch, onExportLayout, clearChart, importedLayout, removeAllComparisons, isAnimationEnabled = true, showLastDigitStats = false, scrollToEpoch, scrollToEpochOffset = 0, zoom, chartControlsWidgets }) {
         this.chartId = id;
         this.settings = settings;
         this.isConnectionOpened = isConnectionOpened;
+        this.chartStatusListener = chartStatusListener;
+        this.status = 'loading';
```
morteza-binary: @nazaninreihani It'd be better to use a boolean instead of a string since there are only two states ('ready` and `loading`).

src/store/ChartState.js
```diff
@@ -174,6 +184,7 @@ class ChartState {
                 this.restoreDrawings(this.stxx, this.stxx.chart.symbol);
                 if (this.chartStore.loader) {
                     this.chartStore.loader.hide();
+                    this.mainStore.state.setChartStatus('ready');
```
```suggestion
                    this.setChartStatus('ready');
```

src/store/ChartStore.js
```diff
@@ -614,10 +615,12 @@ class ChartStore {
     @action.bound newChart(symbolObj = this.currentActiveSymbol, params) {
         this.stxx.chart.symbolDisplay = symbolObj.name;
         this.loader.show();
+        this.mainStore.state.setChartStatus('loading');
```
morteza-binary: @nazaninreihani It'd be more efficient to invoke `setChartStatus` in `loader.show()` and `loader.hide()` functions instead of replicating it.

nazaninreihani: we call loader.hide() conditionally and that is if loader exists, what if chart loads in a pace that no loader is rendered? :-D then the chart will not send the status

src/store/ChartState.js
```diff
@@ -40,10 +41,12 @@ class ChartState {
         this.chartStore.feed.onPagination(this.setOnPagination.bind(this));
     };
 
-    @action.bound updateProps({ id, settings, isConnectionOpened, symbol, granularity, chartType, startEpoch, endEpoch, onExportLayout, clearChart, importedLayout, removeAllComparisons, isAnimationEnabled = true, showLastDigitStats = false, scrollToEpoch, scrollToEpochOffset = 0, zoom, chartControlsWidgets }) {
+    @action.bound updateProps({ id, settings, isConnectionOpened, chartStatusListener, symbol, granularity, chartType, startEpoch, endEpoch, onExportLayout, clearChart, importedLayout, removeAllComparisons, isAnimationEnabled = true, showLastDigitStats = false, scrollToEpoch, scrollToEpochOffset = 0, zoom, chartControlsWidgets }) {
         this.chartId = id;
         this.settings = settings;
         this.isConnectionOpened = isConnectionOpened;
+        this.chartStatusListener = chartStatusListener;
+        this.chartStatus = false;
```
```suggestion
        this.isChartLoaded = false;
```
```suggestion
        this.isChartReady = false;
```

PR #737, opened 2019-04-25
Markers on top

shayanbinary:


app/test.jsx
```diff
@@ -161,8 +173,9 @@ class App extends Component {
             granularity,
             isConnectionOpened: true,
             highLow: {},
+            barrierType: 'OUTSIDE',
             draggable: true,
-            markers: [],
+            markers,
```
shayanbinary: this modification is just for this test, perhaps we remove it later

app/test.jsx
```diff
@@ -283,16 +296,24 @@ class App extends Component {
         this.setState({ hidePriceLines: evt.target.checked });
     };
 
-    onColorChange =(evt) => {
+    onShadeColorChange = (evt) => {
```
shayanbinary: also, I made some modification to `test.jsx` which binary2 needed and also make QA test easy

src/components/Chart.jsx
```diff
@@ -72,15 +72,17 @@ class Chart extends Component {
...
             position,
             showLastDigitStats,
+            yAxiswidth,
         } = this.props;
 
         const currentPosition = `cq-chart-control-${(chartControlsWidgets && position && !isMobile) ? position : 'bottom'}`;
         const contextWidth =  !isMobile ? `smartcharts-${containerWidth}` : '';
         const TopWidgets = topWidgets || this.defaultTopWidgets;
         const BottomWidgets = !bottomWidgets && showLastDigitStats ? LastDigitStats : bottomWidgets;
+        const HasMarkers = children.length ? 'smartcharts--has-markers' : '';
```
shayanbinary: if there are any markers, then increase the `subholder` z-index

src/store/BarrierStore.js
```diff
@@ -114,22 +114,26 @@ export default class BarrierStore {
...
         // barrier 1 cannot go below barrier 2
         this._high_barrier.priceConstrainer = (newPrice) => {
+            let nextPrice = newPrice;
             if (this._low_barrier.visible) {
                 if (newPrice < this._low_barrier.realPrice) {
-                    return this._high_barrier.realPrice;
+                    nextPrice = this._high_barrier.realPrice;
                 }
             }
+            this.mainStore.chart.calculateYaxisWidth(nextPrice);
```
shayanbinary: on each barriers update price, we calculate the width of barriers' price and update `y-axis`

src/store/ChartStore.js
```diff
@@ -314,6 +317,15 @@ class ChartStore {
             }
             let yaxisLabelStyle = this.yaxisLabelStyle;
             if (yax.yaxisLabelStyle) yaxisLabelStyle = yax.yaxisLabelStyle;
+
+            // try to place price label in the y-axis
+            width -= 16;
+            x += 24;
```
shayanbinary: current price label has a triangle on the left hand, these constant number used for that.

src/store/ChartStore.js
```diff
@@ -623,7 +648,10 @@ class ChartStore {
                 return;
             }
             this.state.restoreDrawings();
+            const { Close } = this.context.stx.currentQuote();
+            this.calculateYaxisWidth(Close);
         };
+        this.yAxiswidth = 0;
```
shayanbinary: on each `symbol/market` change, we re-calculate the `y-axis` width due to number

src/chartiq_injections/plotterDrawText.js
```diff
@@ -0,0 +1,21 @@
+export const plotterDrawText = () => {
```
morteza-binary: @shayanbinary Would you please add some comments why the `drawText` has been completely overwritten instead of using injection for further references?

src/components/Chart.jsx
morteza-binary: @shayanbinary would you please add this comment to the code?

```suggestion
        // if there are any markers, then increase the subholder z-index
        const HasMarkers = children.length ? 'smartcharts--has-markers' : '';
```

src/store/ChartStore.js
```diff
@@ -275,6 +278,8 @@ class ChartStore {
         this.chartNode = this.rootNode.querySelector('.ciq-chart-area');
         this.chartControlsNode = this.rootNode.querySelector('.cq-chart-controls');
 
+        CIQ.Plotter.prototype.yAxiswidth = () => this.yAxiswidth;
```
morteza-binary: @shayanbinary it'd be better to rename it to `getYAxixWidth` to make explicit that it's a function.

src/store/ChartStore.js
```diff
@@ -297,11 +302,20 @@ class ChartStore {
                 }
             } catch (e) { width = yax.width; } // Firefox doesn't like this in hidden iframe
 
-            let x = this.width - width;
+            width -= 14;
```
morteza-binary: @shayanbinary would you please comment what is `14` that is deducted from `width`?

src/store/ChartStore.js
```diff
@@ -315,6 +329,12 @@ class ChartStore {
             }
             let yaxisLabelStyle = this.yaxisLabelStyle;
             if (yax.yaxisLabelStyle) yaxisLabelStyle = yax.yaxisLabelStyle;
+
+            if (this.labelType !== 'crosshair' && this.labelType !== 'countdown') {
+                x += 14;
```
morteza-binary: @shayanbinary ditto

PR #743, opened 2019-05-01
Morteza/fix_status_listener

morteza-binary:
Added a check to verify the statusListnere function is passed before invoking

### DIFF
--- a/src/store/ChartState.js
+++ b/src/store/ChartState.js
@@ -175,7 +175,9 @@ class ChartState {
     @action.bound setChartIsReady(isChartReady) {
         if (this.isChartReady !== isChartReady) {
             this.isChartReady = isChartReady;
-            this.chartStatusListener(isChartReady);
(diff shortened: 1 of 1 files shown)


src/store/ChartState.js
```diff
@@ -175,7 +175,9 @@ class ChartState {
     @action.bound setChartIsReady(isChartReady) {
         if (this.isChartReady !== isChartReady) {
             this.isChartReady = isChartReady;
-            this.chartStatusListener(isChartReady);
+            if (this.chartStatusListener || typeof this.chartStatusListener === 'function') {
```
MahboobehMohammadi: @morteza-binary shouldn't it be `&&`?

morteza-binary: yes, sorry

```suggestion
            if (this.chartStatusListener && typeof this.chartStatusListener === 'function') {
```

PR #746, opened 2019-05-02
fix chart design issues

nazaninreihani:


sass/components/_ciq-list.scss
```diff
@@ -16,10 +16,10 @@
     }
     .ciq-list-item {
         cursor: pointer;
-        line-height: 32px;
+        line-height: normal;
         font-size: 1.2em;
-        height: 32px;
-        padding: 0 0 0 16px;
+        height: auto;
+        padding: 8px 16px;
```
morteza-binary: Regarding the design, it should be like this. 
```suggestion
        padding: 8px 0 8px 16px;
```

PR #751, opened 2019-05-06
Mahboobeh/Chart animation

MahboobehMohammadi:


src/store/ChartState.js
```diff
@@ -39,20 +39,19 @@ class ChartState {
         this.stxx.addEventListener('layout', this.saveLayout.bind(this));
         this.stxx.addEventListener('symbolChange', this.saveLayout.bind(this));
         this.stxx.addEventListener('drawing', this.saveDrawings.bind(this));
-        this.stxx.addEventListener('move', this.scrollListener.bind(this));
+        // this.stxx.addEventListener('move', this.scrollListener.bind(this));
```
morteza-binary: @MahboobehMohammadi By commenting out this line, the chart doesn't show the lines between spots when scrolling the chart to left or right.

MahboobehMohammadi: @morteza-binary causing it will keep this issue. i think it has to be rewrite based on new changes in this PR, but i don't know how you set `maxtick` here so i just comment it here so you can check it after this is merged.
image

PR #752, opened 2019-05-07
adjust the barrier top, according to chart height

shayanbinary:


src/store/BarrierStore.js
```diff
@@ -134,6 +134,7 @@ export default class BarrierStore {
     get context() { return this.mainStore.chart.context; }
     get stx() { return this.context.stx; }
     get chart() { return this.stx.chart; }
+    get chartHeight() { return this.mainStore.chart.chartHeight; }
```
morteza-binary: @shayanbinary it seems it's not used anymore.

PR #758, opened 2019-05-08
Morteza/chart_loading_issue

morteza-binary:
Fixed the issue loading the chart when there aren't any markers


morteza-binary:
Duplicate of #754

PR #760, opened 2019-05-08
fix overlapping markers

kellybinary:
- set bottom value for start/end markers if there's a bottom widget

### DIFF
--- a/src/store/MarkerStore.js
+++ b/src/store/MarkerStore.js
@@ -161,7 +161,7 @@ export default class MarkerStore {
         // Y axis positioning logic
         if (this.yPositioner.toLowerCase() === 'none') {
             // set the bottom value if there's a bottomWidget, else set it to undefined
-            this.bottom = this.stx.chart.yAxis.initialMarginBottom !== 100 ? 125 : undefined;
+            this.bottom = this.stx.chart.yAxis.initialMarginBottom === 200 ? 125 : 20;
             this.showMarker();
             return;
         }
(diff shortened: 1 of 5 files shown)


src/store/MarkerStore.js
```diff
@@ -160,7 +160,8 @@ export default class MarkerStore {
 
         // Y axis positioning logic
         if (this.yPositioner.toLowerCase() === 'none') {
-            this.bottom = undefined;
+            // set the bottom value if there's a bottomWidget, else set it to undefined
+            this.bottom = this.stx.chart.yAxis.initialMarginBottom !== 50 ? 130 : undefined;
```
morteza-binary: @kellybinary it'd be better to use `this.stx.chart.yAxis.initialMarginBottom` to set `this.bottom` value.

kellybinary: @morteza-binary 
Screenshot 2019-05-08 at 16 27 47

using ```this.stx.chart.yAxis.initialMarginBottom``` as `this.bottom` value will cause the the end marker line overlapping with the spot too

src/store/MarkerStore.js
```diff
@@ -160,7 +160,8 @@ export default class MarkerStore {
 
         // Y axis positioning logic
         if (this.yPositioner.toLowerCase() === 'none') {
-            this.bottom = undefined;
+            // set the bottom value if there's a bottomWidget, else set it to undefined
+            this.bottom = this.stx.chart.yAxis.initialMarginBottom !== 50 ? 125 : undefined;
```
morteza-binary: @kellybinary the default margin has been changed to `100` in my last PR #762. So it should be like this
```suggestion
            this.bottom = this.stx.chart.yAxis.initialMarginBottom !== 100 ? 125 : undefined;
```

PR #777, opened 2019-05-22
kelly/endline_issue

kellybinary:


src/feed/Feed.js
```diff
@@ -282,7 +282,7 @@ class Feed {
         const subscription = this._activeStreams[key];
         if (!subscription) { return; }
         const lastEpoch = subscription.lastStreamEpoch;
-        if (this.endEpoch && lastEpoch > this.endEpoch) {
+        if (this.endEpoch && lastEpoch >= this.endEpoch) {
```
morteza-binary: @kellybinary I don't think it fixes the issue since the lastEpoch is always bigger that endEpoch when there isn't any tick for endEpoch.

PR #778, opened 2019-05-22
Limit indicators

nazaninreihani:


src/store/StudyLegendStore.js
```diff
@@ -88,10 +88,36 @@ export default class StudyLegendStore {
...
-        logEvent(LogCategories.ChartControl, LogActions.Indicator, `Add ${item}`);
-        this.menu.setOpen(false);
+        if (this.stx.layout && Object.keys(this.stx.layout.studies || []).length < 5) {
+            const sd = CIQ.Studies.addStudy(this.stx, item);
+            this.changeStudyPanelTitle(sd);
+            logEvent(LogCategories.ChartControl, LogActions.Indicator, `Add ${item}`);
+            this.menu.setOpen(false);
+        } else {
+            this.mainStore.notifier.notify({
+                text: t.translate('You can\'t have more that than 5 open Indicators.'),
```
kellybinary: i think you can remove the word `that`

src/store/StudyLegendStore.js
```diff
@@ -88,10 +88,36 @@ export default class StudyLegendStore {
...
+            });
+            this.mainStore.state.setShouldMinimiseLastDigit(true);
+        } else {
+            Object.keys(CIQ.Studies.studyLibrary).forEach((st) => {
+                if (!CIQ.Studies.studyLibrary[st].overlay) {
+                    CIQ.Studies.studyLibrary[st].panelHeight = null;
+                }
+            });
+            this.mainStore.state.setShouldMinimiseLastDigit(false);
+        }
```
kellybinary: i think can be simplified:
```
const should_minimise_last_digit = Object.keys(this.stx.panels).length > 2;
Object.keys(CIQ.Studies.studyLibrary).forEach((st) => {
   if (!CIQ.Studies.studyLibrary[st].overlay) {
      CIQ.Studies.studyLibrary[st].panelHeight = should_minimise_last_digit ? 80 : null;
   }
});
this.mainStore.state.setShouldMinimiseLastDigit(should_minimise_last_digit);
```

src/store/StudyLegendStore.js
nazaninreihani: aw I did not see that, thanks

src/store/StudyLegendStore.js
```diff
@@ -91,10 +91,28 @@ export default class StudyLegendStore {
...
+            this.menu.setOpen(false);
+        } else {
+            this.mainStore.notifier.notify({
+                text: t.translate('You can\'t have more than 5 open Indicators.'),
+            });
+        }
+    }
+
+    // Temporary prevent user from adding more than 5 indicators
+    // All traces can be removed after new design for studies
```
morteza-binary:
```suggestion
    // TODO All traces can be removed after new design for studies
```

src/store/ChartState.js
```diff
@@ -23,6 +23,7 @@ class ChartState {
     @observable isOnPagination = false;
     @observable paginationEndEpoch;
     @observable isChartClosed = false;
+    @observable shouldMinimiseLastDigits = false;
```
morteza-binary: @nazaninreihani, while `shouldMinimiseLastDigits` is a requirement of `LastDigitStats` it would be better to move it to `LastDigitStatsStore` as a computed property.

src/components/categoricaldisplay/ResultsPanel.jsx
```diff
@@ -13,25 +13,41 @@ function getItemCount(category) {
     return count;
 }
 
-const CategoryTitleClassName = (categoryId, activeHeadKey, activeHeadTop) => {
+const CategoryTitleClassName = (categoryId, activeHeadKey, activeHeadTop, categorySubtitle) => {
```
morteza-binary: @nazaninreihani if `categorySubtitle` is a boolean variable, it'd be better to name in a way that conveys its value. e.g `hasCategorySubtitle`, `hasSubtitle`, ...

PR #790, opened 2019-05-28
shayan/change color of dialog header and footer

shayanbinary:


sass/components/_ciq-chart-setting.scss
```diff
@@ -13,6 +13,9 @@
         > .icon-back-menu {
             display: inline-block;
             cursor: pointer;
+            @include themify($themes) {
+                color: themed('ChartSettingHeadText');
```
nazaninreihani: I can't find this color in themes.scss file, which line is it in?

nazaninreihani: `ChartSettingHeadText`

shayanbinary: oops, it should be `DialogSettingHeadText`.

PR #798, opened 2019-05-30
Do not allow whitespace on the right after chart reaches the end of data

nazaninreihani:


src/store/ChartState.js
```diff
@@ -501,6 +501,15 @@ class ChartState {
         if (grab && this.stxx.chart.lockScroll) {
             this.stxx.chart.lockScroll = false;
         }
+        if (grab && this.stxx && this.stxx.chart) {
+            const dataSegment = this.stxx.chart.dataSegment;
+            if (dataSegment) {
+                const startPointOnChartData = dataSegment[0];
+                const barsDisplayedOnScreen = Math.floor(this.stxx.chart.width / this.stxx.layout.candleWidth);
```
morteza-binary: @nazaninreihani You can use `this.stxx.chart.maxTicks`. Its value is as same as `barsDisplayedOnScreen` and it shows the number of bars(ticks) on the screen.

src/feed/Feed.js
```diff
@@ -272,6 +275,12 @@ class Feed {
         }
     }
 
+    setHasReachedEndOfData(hasReachedEndOfData) {
```
morteza-binary: @nazaninreihani You can use `Feed.Event_ON_PAGINATION` emitter to set `hasReachedEndOfData` in the corresponding store(`stateStore`).

nazaninreihani: We need to set hasReachedEndOfData to false when starting to fetch a new feed, we can't do that in Feed.Event_ON_PAGINATION, If I want to this I should add an event for starting new feed, what do you think?

src/store/ChartState.js
```diff
@@ -253,6 +253,8 @@ class ChartState {
     }
 
     @action.bound setOnPagination({ end }) {
+        // TODO: remove next line when the fix of markers is done
```
nazaninreihani: @kellybinary this is the temporary fix that needs to be removed later

kellybinary: this fixes the missing chart loader, thanks @nazaninreihani  🎉

src/store/ChartStore.js
```diff
@@ -562,6 +562,10 @@ class ChartStore {
         }
     };
 
+    getMaxMasterDataSize(granularity) {
+        return granularity === 86400 ? Math.floor(2.8 * 365) : 5000;
```
morteza-binary: @nazaninreihani how about 8 hours granularity? the number of ticks is less than 5000 for 3 years.

nazaninreihani: oh okay I'll add that too

PR #802, opened 2019-06-03
kelly/fix_scroll_to_left

kellybinary:


src/store/MarkerStore.js
```diff
@@ -30,6 +30,11 @@ export default class MarkerStore {
         this._injectionId = this.stx.prepend('positionMarkers', this.updateMarkerTick);
     }
 
+    @computed get shouldDrawMarkers() {
+        // Checks if the chart has been scrolled to the leftmost of the screen
+        return this.stx.chart.isScrollLocationChanged;
```
morteza-binary: @kellybinary Also we can check that the masterData/dataSet has valid values or not here to prevent markers from to being drawn when the chart is empty.

kellybinary: @morteza-binary it is already added at line 73 😄

PR #803, opened 2019-06-05
Shayan/find the latest data that contain close value

shayanbinary:


src/store/ChartStore.js
```diff
@@ -665,7 +665,10 @@ class ChartStore {
             if (currentQuote && currentQuote.Close) {
                 this.calculateYaxisWidth(currentQuote.Close);
             } else {
-                this.calculateYaxisWidth(this.stxx.masterData.slice(-1).Close);
+                const lastDataWitClose = this.stxx.masterData
+                    .filter(x => x.Close).slice(-1);
```
morteza-binary: @shayanbinary It'd be better to use `find` instead of `filter` since `find` is faster than 'filter'
```suggestion
                    .find(x => x.Close);
```

src/store/ChartStore.js
```diff
@@ -665,7 +665,10 @@ class ChartStore {
...
+
+                this.calculateYaxisWidth(lastDataWitClose[0].Close);
```
morteza-binary:
```suggestion
                this.calculateYaxisWidth(lastDataWitClose.Close);
```

src/store/ChartStore.js
shayanbinary: @morteza-binary here we need the last item that has, valid `Close` prop. `find` return the first item.

morteza-binary: @shayanbinary we just need the width of the price, there's not any difference between the first and the last one.

PR #807, opened 2019-06-14
Shayan/The chart animation in purchase mode is different from the trading mode 

shayan-binary:


shayan-binary-2:
@morteza-binary Please close this PR. we did thin on another PR.

shayan-binary-2:
it's out date and we implement this in another card

PR #810, opened 2019-06-14
Shayan/Chart Table Type is not readable in the light theme

shayan-binary:


morteza-binary:
@shayanbinary Please solve the conflict.

PR #814, opened 2019-06-24
Cannot open the table type for the second time

shayan-binary:


shayan-binary-2:
@zaki-binary please close this PR, it's for long time ago, for now, we do not use table anymore

PR #823, opened 2019-07-09
kelly/refactor updateProps in chartState.js

kellybinary:


src/store/ChartState.js
```diff
@@ -75,101 +76,70 @@ class ChartState {
...
+        this.setChartType(chartType);
+        this.setChartGranularity(granularity);
+
+        switch (true) {
+        case (symbol !== this.symbol): {
+            this.symbol = symbol;
             if (this.mainStore.chart && this.mainStore.chart.feed) {
                 this.mainStore.chart.feed.onMasterDataUpdate(this.scrollChartToLeft);
             }
+            break;
```
morteza-binary: @kellybinary is the `break` in each case intentionally? while one the case break the switch other changes won't be applied to the chart until the next update.

src/store/ChartState.js
```diff
@@ -75,101 +76,70 @@ class ChartState {
...
+        this.showLastDigitStats = showLastDigitStats;
 
-        if (this.symbol !== symbol) {
-            this.symbol = symbol;
+        this.startEpoch = startEpoch;
+        this.endEpoch = endEpoch;
+        this.margin = margin;
 
+        // IMPORTANT! Set chartType first then set chart granularity
+        this.setChartType(chartType);
```
morteza-binary: @kellybinary we can move these to function into the switch's body but without break.

src/store/ChartState.js
kellybinary: @morteza-binary yes, the `break` are intentional, we don't need to check/execute every condition for any props changes

src/store/ChartState.js
kellybinary: @morteza-binary sure 👍

aminroosta:
@kellybinary 

```js
switch(true) {
    case (condition) : { 
        break;
     }
     case (condition2) : {
        break;
     }
}
```

this is too creative, it took me a minute to figure out what's go on.
I was surprised to see a `case` on a Boolean.

 I would suggest to revert to `if() { } else if() { } ...` syntax.

PR #824, opened 2019-07-10
shayan/navigation widget

shayan-binary:


src/store/NavigationWidgetStore.js
```diff
@@ -0,0 +1,43 @@
...
+     @action.bound onScale() {
+         let scrollToDate = null;
+
+         if (!this.stateStore.scrollToEpoch) {
+             const { dataSet } = this.stxx.chart;
+             if (dataSet && dataSet.length) {
+                 scrollToDate = getUTCEpoch(dataSet[0]);
+             }
+         }
+         this.stateStore.scrollChartToLeft(scrollToDate);
```
kellybinary: better to call this function `onMasterDataUpdate` to prevent chart from getting scrolled off screen

src/store/ChartState.js
```diff
@@ -393,29 +395,14 @@ class ChartState {
...
-            0, scrollToTarget);
-            this.stxx.draw();
+
+            const tick = this.stxx.tickFromDate(startEntry.DT);
+            const startTick = this.stxx.chart.dataSet.length - tick;
+            this.stxx.chart.lockScroll = true;
+            this.stxx.chart.isScrollLocationChanged = true; // set to true to draw markers
+
+            if (!this.startEpoch) this.scrollChartToLeftAnimator(startTick, startEntry);
+            else this.scrollChartToLeftGraph(startTick, startEntry);
```
kellybinary: @shayanbinary is this `else` clause necessary?  shouldn't the `scrollChartToLeftAnimator` takes care of it? i think charts need to wait until it gets `startEpoch` value to start the scrolling animation

morteza-binary:
Duplicate of #731

PR #826, opened 2019-07-16
Morteza/movement_animation

morteza-binary:


kelly-binary:
@morteza-binary LGTM

sass/_themes.scss
```diff
@@ -420,7 +421,8 @@ $themes: (
         HighestLowestMarkerLabelBg: $COLOR_DARK_GRAY_2,
         HighestLowestMarkerLabelText: $COLOR_WHITE,
         HighestSpotColor: $COLOR_GREEN_1,
-        LowestSpotColor: $COLOR_RED
+        LowestSpotColor: $COLOR_RED,
+        ChartBackgroundOpacity: 0.4
```
brandon-binary: The `ChartBackgroundOpacity` for white theme is 0.5, is this suppose to be 0.4?

sass/styles/chart.scss
```diff
@@ -761,6 +761,28 @@ html, body {
...
+        @include themify($themes) {
+            opacity: themed('ChartBackgroundOpacity');
+        }
+        background-image: url('https://static.cdnpub.info/traderoom/map.png'); // TODO the image's url should be changed after getting new image from UI/UX team.
+        background-repeat: no-repeat;
+        background-position: 50% 0;
+        -ms-background-size: cover;
+        -o-background-size: cover;
+        -moz-background-size: cover;
+        -webkit-background-size: cover;
```
brandon-binary: just out of curiosity, isn't there a plugin that add all the vendor prefixes for us? Why are we not using it?

src/components/ui/Animation.js
```diff
@@ -323,4 +228,50 @@ export default function animateChart(stx, animationParameters, easeMachine) {
...
+            if (stx.animations.liveScroll.running) {
+                stx.animations.liveScroll.stop();
+            }
+
+            stx.animations.liveScroll.run((bar) => {
+                this.micropixels = -bar;
+                this.draw();
+            }, 0, stx.layout.candleWidth);
+        } else if (!stx.isHome() || stx.chart.lockAutoScroll) {
+            if (stx.animations.liveScroll) {
```
kelly-binary: should this check be `stx.animations.liveScroll.running` ?

src/store/ChartStore.js
```diff
@@ -451,6 +451,7 @@ class ChartStore {
         manageElement.textContent = t.translate('right-click to manage');
 
         if (this.state.isAnimationEnabled) animateChart(stxx, { stayPut: true }, tickAnimator);
+        stxx.chart.lockScroll = true;
```
kelly-binary: should this be `lockAutoScroll`?

src/components/ui/Animation.js
```diff
@@ -52,18 +60,20 @@ export default function animateChart(stx, animationParameters, easeMachine) {
...
+        stx.chart.granularity = calculateGranularity(this.layout.interval, this.layout.timeUnit);
 
-    function initMarketSessionFlags() {
-        filterSession = false;
-        nextBoundary = null;
-    }
+        if (stx.chart.granularity === 0) {
+            stx.chart.lockScroll = true;
+        } else {
+            stx.chart.lockScroll = false;
```
kelly-binary: ditto, should this be `lockAutoScroll`?

src/store/ChartState.js
```diff
@@ -372,37 +394,33 @@ class ChartState {
...
-                } else {
-                    this.stxx.chart.scroll = scroll;
-                }
-            },
-            0, scrollToTarget);
+            this.stxx.micropixels = 0;
             this.stxx.draw();
+            this.stxx.chart.entryTick = this.stxx.tickFromDate(startEntry.DT); // the calculation of entry tick should be done after draw
+
+            const scrollToTarget = this.stxx.chart.dataSet.length - this.stxx.chart.entryTick;
```
kelly-binary: this.stxx.scrollTo doesn't seem to work when `scrollToTarget` is 1. 
suggest to keep this a constant of `2` instead :)

src/store/ChartState.js
```diff
@@ -97,14 +97,24 @@ class ChartState {
...
+            this.exportLayout();
+        }
+
+
         if (chartType !== this.chartType && this.context) {
             this.setChartType(chartType);
-        } else if (granularity !== undefined && granularity !== this.granularity) {
+        }
+
+        if (granularity !== undefined && granularity !== this.granularity) {
```
kelly-binary: any reason changing to `if` instead of `else if`?

src/store/ChartStore.js
morteza-binary: No, here it should be `lockScroll` to prevent ChartIQ from autoscrolling.

sass/_themes.scss
morteza-binary: yes, it's correct to make it more visible in white theme. Anyway this background should be changed after the new one is done by the UI/UX team.

sass/styles/chart.scss
morteza-binary: I don't think there are any plugins for it in SmartChart project, I should check it again.

src/components/ui/Animation.js
morteza-binary: here Also, it should be `lockScroll`

src/store/ChartState.js
morteza-binary: @kelly-binary yes, because it should be possible to change chartType and granularity at the same time.

PR #828, opened 2019-07-16
Nazanin/Stream issue

nazanin-binary:


src/binaryapi/BinaryAPI.js
```diff
@@ -50,6 +51,10 @@ export default class BinaryAPI {
         return this.requestForget(request, callback);
     }
 
+    forgetStream(subscription_id) {
+        return this.requestForgetStream(subscription_id);
```
nazanin-binary: `        if(this.requestForgetStream && typeof this.requestForgetStream === 'function') {
return this.requestForgetStream(subscription_id);
 }
`

PR #830, opened 2019-07-17
Nazanin/show markers and current spot when it's ready

nazanin-binary:


sass/components/_ciq-crosshair.scss
```diff
@@ -38,7 +38,8 @@
             color: themed('CrosshairText');
             background: themed('CrosshairBg');
         }
-        display: flex;
+        display: none;
+        // we make it display flex in js
```
morteza-binary: @nazanin-binary Regarding the changes in crosshair display, IMO showing and hiding the crosshair by setting `display` to `none` has some rendering cost. 
@aminroosta Would you please check it?

aminroosta: @nazanin-binary changing `display` will force browser to re calculate the css layout.
It's expensive and it's better to avoid.

The alternative is to set `visibility: hidden` or `translate: transform  (-5000px, -5000px)` or `opacity: 0` depending on what we are trying to achieve.

src/components/Chart.jsx
```diff
@@ -113,7 +114,7 @@ class Chart extends Component {
                                         }
                                     </RenderInsideChart>
                                     <RenderInsideChart at="subholder">
-                                        {children}
+                                        {isChartReady && children}
```
aminroosta: `RenderInsideChart` should and does take care of it's children. 
Only when the chart is ready we should render them, if that's not the case it needs to be fixed in `RenderInsideChart`.

src/store/CrosshairStore.js
```diff
@@ -354,7 +355,7 @@ class CrosshairStore {
 
     // YES! we are manually patching DOM, Because we don't want to pay
     // for react reconciler & mox tracking observables.
-    updateTooltipPosition({ top, left, rows }) {
+    updateTooltipPosition({ top, left, rows, show = false }) {
```
aminroosta: pass `top: -5000px, left: -5000px` instead, similar to how we do it here
https://github.com/binary-com/SmartCharts/pull/830/files#diff-2f67da90737005346bff0e111831096cR94

src/store/MarkerStore.js
```diff
@@ -13,7 +13,7 @@ export default class MarkerStore {
     children;
     x;
     y;
-    @observable display;
+    @observable display = 'none';
```
aminroosta: MarkerStore is deprecated, we will remove it in next release.
Deriv.app has already migrated to the new FastMarker.

sass/components/_ciq-crosshair.scss
nazanin-binary: I removed all css changes

nazanin-binary:
https://github.com/binary-com/SmartCharts/pull/836

PR #832, opened 2019-07-18
brandon/cbc

brandon-binary:
add a check to make sure chart is only cleaned when a chart is ready

### DIFF
--- a/src/store/ChartState.js
+++ b/src/store/ChartState.js
@@ -407,21 +407,21 @@ class ChartState {
     }
 
     cleanChart() {
-        if (this.clearChart && this.isChartReady) {
-            // Remove comparsions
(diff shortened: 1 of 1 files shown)


src/store/ChartState.js
```diff
@@ -407,20 +407,21 @@ class ChartState {
...
-        for (const field in this.stxx.chart.series) {
-            this.stxx.removeSeries(field);
-        }
-        // Remove indiactors
-        for (const id in this.stxx.layout.studies) {
-            const sd = this.stxx.layout.studies[id];
-            CIQ.Studies.removeStudy(this.stxx, sd);
-        }
-        this.stxx.clearDrawings();
+        if (this.clearChart && this.isChartReady) {
```
negarn: why not keep the return?

PR #836, opened 2019-07-23
Nazanin/Markers after chart is ready

nazanin-binary:


src/components/RenderInsideChart.jsx
```diff
@@ -25,7 +25,12 @@ class RenderInsideChart extends Component {
         });
     }
 
+    componentDidMount() {
+        reaction(() => this.props.isChartReady, () => this.forceUpdate());
```
morteza-binary: @nazanin-binary Does it need to use `reaction` and `forceUpdate` while `isChartReady` is observable and part of the component property?

nazanin-binary: right it's not needed, will remove it, thanks

src/components/RenderInsideChart.jsx
```diff
@@ -26,6 +26,8 @@ class RenderInsideChart extends Component {
     }
 
     render() {
+        if (!this.props.isChartReady) return (null);
```
morteza-binary:
```suggestion
        if (!this.props.isChartReady) return(null);
```

src/store/ChartState.js
```diff
@@ -411,6 +417,7 @@ class ChartState {
             this.stxx.scrollTo(this.stxx.chart, scrollToTarget, () => {
                 this.stxx.setMaxTicks(5);
                 this.stxx.micropixels = 0;
+                this.setIsChartScrollingToEpoch(false);
                 this.stxx.chart.isScrollLocationChanged = true; // set to true to draw markers
```
morteza-binary: @nazanin-binary it seems that we don't need it anymore since the markers are drawn after scrolling. Actually, after using `FastMarker` by Deriv-app this flag is not usable anymore.

src/store/ChartState.js
```diff
@@ -390,17 +396,17 @@ class ChartState {
...
         if (this.scrollToEpoch && !this.startEpoch) {
+            const dataSetLength = this.stxx.chart.dataSet.length;
             const startEntry = this.stxx.chart.dataSet
                 .find(entry =>  entry.DT.valueOf() === CIQ.strToDateTime(getUTCDate(this.scrollToEpoch)).valueOf());
 
             if (startEntry) {
                 this.stxx.chart.entryTick = this.stxx.tickFromDate(startEntry.DT);
             } else {
-                this.stxx.chart.entryTick = this.stxx.chart.dataSet.length - 1;
+                this.stxx.chart.entryTick = dataSetLength - 1;
```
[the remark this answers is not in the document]
nazanin-binary: yes you're right

PR #839, opened 2019-07-30
Shayan/SmartChart not loaded in binary bot when language change to Espanyol

shayan-binary:
…rring to it


src/components/ChartSetting.jsx
```diff
@@ -47,156 +47,165 @@ const ThemeToggle = ({
...
-                    <div className="ciq-list-item">
-                        <span className="ciq-icon-text">{t.translate('Auto Scale')}</span>
-                        <div className="ciq-action">
-                            <Switch
-                                value={isAutoScale}
-                                onChange={setAutoScale}
-                            />
+class ChartSetting extends Component {
+    componentDidMount() {
+        const { init, activeLanguages } = this.props;
```
morteza-binary: @shayan-binary What happens by changing `activeLanguages` after mounting the component. Assume the page is loaded completely then the value of `activeLanguages` should be changed.

shayan-binary: @morteza-binary so you mean. `activeLanguages` may be changed during the rendering?

shayan-binary: of course, we can make it in a way to update per `activeLanguages` changed, I just thought it's going to initial in the beginning

morteza-binary: @shayan-binary it'd be better to support it too.

src/store/ChartSettingStore.js
```diff
@@ -104,37 +104,67 @@ export default class ChartSettingStore {
...
-               historical                  : this.historical,
-               language                    : this.language.key,
-               position                    : this.position,
-               isAutoScale                 : this.isAutoScale,
-               isHighestLowestMarkerEnabled: this.isHighestLowestMarkerEnabled,
-               theme                       : this.theme,
-           });
-       }
-   }
+    @action.bound updateProps(activeLanguages) {
```
morteza-binary: @shayan-binary it should be like this
`@action.bound updateProps({ activeLanguages }) {`

PR #840, opened 2019-07-30
upgrade_packages

brandon-binary:


package.json
```diff
@@ -50,7 +50,7 @@
     "@babel/register": "^7.0.0",
     "babel-eslint": "^10.0.1",
     "babel-loader": "^8.0.2",
-    "babel-plugin-inline-react-svg": "^1.0.0",
+    "babel-plugin-inline-react-svg": "^1.1.0",
```
morteza-binary: @brandon-binary How about other vulnerability packages

package.json
```diff
@@ -133,5 +133,8 @@
     "react": "^16.4.0",
     "react-dom": "^16.4.0",
     "react-transition-group": "^2.5.0"
+  },
+  "resolutions": {
+    "js-yaml": "^3.13.1"
```
morteza-binary: @brandon-binary would you please specify which package is dependent on `js-yaml`?

brandon-binary: `babel-plugin-inline-react-svg`

package.json
brandon-binary: There shouldn't be other vulnerable packages after upgrading this.

morteza-binary: @brandon-binary After applying your changes in my local machine, `yarn audit` gave me 508 vulnerabilities packages 
image

brandon-binary: Did you reinstall npm_module?

PR #842, opened 2019-08-09
Shayan/Add more space between columns in table type

shayan-fs:


sass/components/_ciq-chart-table.scss
```diff
@@ -16,147 +12,201 @@
...
+                @include themify($themes) {
+                    color: $COLOR_GREEN_1;
+                }
+            }
+            .down {
+                @include themify($themes) {
+                    color: $COLOR_RED;
+                }
+            }
+            &.before-last-child {
```
nazanin-binary: you could use :nth-last-child(2) to get the one before last child if you need it, but here and the following nth-childs, why not using classes like this to prevent indexing? the table is not dynamic, although this is a small component and doesn't have a huge impact so it's your call

sass/components/_ciq-chart-table.scss
```diff
@@ -16,147 +12,201 @@
...
+    tbody {
+        display: block;
+    }
+    .cq-change-cell {
+        display: inline-flex;
+    }
+    &--tick {
+        tr {
+            td, th {
+                &:nth-child(1), &:nth-child(1) {
```
nazanin-binary: can be first-child

src/store/ChartTableStore.js
```diff
@@ -47,44 +54,96 @@ export default class ChartTableStore {
...
+                datas: [data],
+            };
+            this.tableData.unshift(dateGroup);
+        } else {
+            const oldData = dateGroup.datas.find(x => x.Date === dateTime);
+            if (oldData) {
+                let previousClose = 0;
+
+                // if the first group has more than 2 data, let compare
+                if (this.tableData.lenght && this.tableData[0].datas.lenght > 1) {
```
nazanin-binary: there are few typos for `length`

PR #859, opened 2019-09-01
shayan/optimize chart title

shayan-fs:


shayan-fs:
@nazanin-binary it's a mistake

PR #860, opened 2019-09-03
Fix scroll to epoch when comparison chart is available

nazanin-binary:


src/store/ChartState.js
```diff
@@ -167,7 +167,7 @@ class ChartState {
             this.scrollToEpoch = scrollToEpoch;
             if (this.mainStore.chart && this.mainStore.chart.feed && !isSymbolChanged && !isGranularityChanged) {
                 this.setIsChartScrollingToEpoch(true);
-                this.scrollChartToLeft();
+                this.mainStore.chart.feed.onMasterDataUpdate(this.scrollChartToLeft);
```
morteza-binary: @nazanin-binary There is an issue with symbols that have a long delay between two ticks. e.g. for indices symbols since there is a long delay between ticks the chart doesn't scroll to left immediately.

Anyway, in the new design, the chart doesn't need to scroll to left after purchasing a new contract.

PR #863, opened 2019-09-16
Shayan/ issue of market selector

shayan-fs:


src/components/categoricaldisplay/FilterPanel.jsx
```diff
@@ -3,7 +3,7 @@ import {
     CategoryIconMap,
 } from '../Icons.jsx';
 
-const Filter = ({ focusedCategoryKey, activeCategoryKey, handleFilterClick, category, isMobile }) => {
+const Filter = React.memo(({ focusedCategoryKey, activeCategoryKey, handleFilterClick, category, isMobile }) => {
```
shayan-fs: the issue doesn't come from here, but as I checked `Profiler` for market change, this component re-render without any reason, so I cached it

PR #865, opened 2019-09-17
Morteza/remove_auto_scroll

morteza-binary:
Removed chart auto-scrolling

### DIFF
--- a/src/components/ui/Animation.js
+++ b/src/components/ui/Animation.js
@@ -282,9 +282,10 @@ export default function animateChart(stx, animationParameters, easeMachine) {
                 if (this.animations.zoom.hasCompleted) {
                     const candleWidth = this.layout.candleWidth;
                     if (chart.scroll <= chart.maxTicks && !chart.lockScroll) {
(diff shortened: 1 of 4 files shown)


src/components/ui/Animation.js
```diff
@@ -223,7 +279,42 @@ export default function animateChart(stx, animationParameters, easeMachine) {
...
             let beginningOffset = 0;
             if (chartJustAdvanced) {
-                if (!this.animations.zoom.hasCompleted) {
+                if (this.animations.zoom.hasCompleted) {
+                    const candleWidth = this.layout.candleWidth;
+                    if (chart.scroll <= chart.maxTicks && !chart.lockScroll) {
+                        while (this.micropixels > 0) { // If micropixels is larger than a candle then scroll back further
+                            chart.scroll++;
+                            this.micropixels -= candleWidth;
+                        }
```
aminroosta:
```
if (this.micropixels > 0) {
    const count = Math.ceil(this.micropixels / candleWidth);
    this.scroll += count;
    this.micropixels -= count * candleWidth;
}
```

Let's remove the `while` since it could be expensive.

PR #866, opened 2019-09-22
shayan/merge importLayout and restoreLayout

shayan-fs:


shayan-fs:
As `deriv-app` no longer use `importedLayout` to build chart so purpose of this PR not longer needed.

PR #876, opened 2019-09-30
Shayan/Tooltip can cover symbol details

shayan-fs:


src/store/ChartTitleStore.js
```diff
@@ -116,4 +117,12 @@ export default class ChartTitleStore {
             }
         }
     }
+
+    @action.bound onMouseEnter() {
```
morteza-binary: @shayan-fs it doesn't need `@action-bound` here
```suggestion
    onMouseEnter = () => {
```

src/store/ChartTitleStore.js
```diff
@@ -116,4 +117,12 @@ export default class ChartTitleStore {
...
+        this.crosshairStore.updateVisibility(false);
+    }
+
+    @action.bound onMouseLeave() {
```
morteza-binary:
```suggestion
    onMouseLeave = () => {
```

PR #878, opened 2019-10-01
shayan/An issue in Fav list in competitors dropdown

shayan-fs:


src/store/CategoricalDisplayStore.js
```diff
@@ -130,7 +131,8 @@ export default class CategoricalDisplayStore {
         if (this.pauseScrollSpy || !this.scrollPanel) { return; }
         if (this.filteredItems.length === 0) { return; }
 
-        const categoryTitleHeight = 40;
+        // hits: 40px for title hight + 4px for content bottom border
+        const categoryTitleHeight = 44;
```
shayan-fs: this is not related to this card. during recent style modification a bottom border added to categorical display section bottom, which cause a jump in fix title.

sass/components/_categorical-display.scss
```diff
@@ -156,8 +156,10 @@
             border: 1px solid themed('CatDisplayCatContentBorder');
         }
 
-        > div:nth-child(2) {
-            border-radius: 4px 4px 0 0;
+        &.has-subcategory {
+            > div:nth-child(2) {
+                border-radius: 4px 4px 0 0;
+            }
```
shayan-fs: On Categorical Display, we have two kinds of sections. one of them has subcategory(Forex, ...) and another one has no sub-category(Favorite, active). For border-radius, first and the last item of each content should have half-radius and the rest should have full-radius(each box corner). I handle it by adding `has-subcategory`

PR #879, opened 2019-10-03
Shayan/Graph broke when the screen resize

shayan-fs:


shayan-binary-2:
@morteza-binary Please close this PR, we handle this on another card.

PR #881, opened 2019-10-10
Shayan/chart type issue

shayan-fs:
the issue comes from the place which application and the state(ChartState) do not update with the new changes after loading template.


app/index.jsx
```diff
@@ -241,14 +235,10 @@ class App extends Component {
         <>
             {isMobile ? '' : <CrosshairToggle />}
             <ChartTypes
-                onChange={(chartType) => {
-                    this.setState({
-                        chartType,
-                    });
-                }}
+                onChange={chartType => this.setState({ chartType })}
```
morteza-binary: @shayan-fs it'd be better to define the functions outside of the JSX tags.

app/index.jsx
```diff
@@ -241,14 +235,10 @@ class App extends Component {
...
             />
             <Timeperiod
-                onChange={this.changeGranularity}
+                onChange={timePeriod => this.setState({ granularity: timePeriod })}
```
morteza-binary: @shayan-fs ditto

src/store/ChartState.js
```diff
@@ -113,7 +113,7 @@ class ChartState {
 
         if (chartType !== this.chartType && this.context) {
             if (chartType === 'table') this.prevChartType = this.chartTypeStore.type.id;
-            this.setChartType(chartType);
+            this.setChartChartType(chartType);
```
morteza-binary: @shayan-fs shouldn't it be `this.setChartType`?
```suggestion
            this.setChartType(chartType);
```

src/store/ViewStore.js
```diff
@@ -87,27 +88,36 @@ export default class ViewStore {
         const stx = this.stx;
 
         const importLayout = () => {
+            // This condition is to make spline chart appear as spline chart
+            // Both line chart and spline chart are of type mountain but with different tensions
```
morteza-binary: @shayan-fs These two lines should be moved to line 94.

src/store/ChartState.js
shayan-fs: this is use to be `setChartType` then I changed it to `setChartChartType` and add `setChartType` to the state. `setChartChartType` change the chart type, `setChartType` change the state store chart-type.

morteza-binary: @shayan-fs but the name is like a typo and make some confusing it'd be better to choose a better name.

shayan-fs: @morteza-binary two option, 1) rename `setChartChartType` to `setCIQChartType`. or rename `setChartType` to `setStateChartType` and  `setChartChartType` to `setChartType`. should notice that by this rule all state function which has this issue renamed so that we have a constant name structure for all methods

src/store/TimeperiodStore.js
```diff
@@ -38,6 +38,8 @@ export default class TimeperiodStore {
         ], this.updateCountdown.bind(this));
 
         this.context.stx.addEventListener('newChart', this.updateDisplay);
+
+        reaction(() => this.mainStore.state.granularity, granularity => this.onGranularityChange(granularity));
```
morteza-binary: @shayan-binary The reaction will cause an exception if `onGranularityChange` is not assigned for `Timeperiod` component. it would be better to check `onGranularityChange` to be a function before invoking it or assign a default value to it at line 22 (e.g `onGranularityChange = () => null`

src/store/ViewStore.js
```diff
@@ -92,6 +92,7 @@ export default class ViewStore {
                 this.mainStore.studies.updateActiveStudies();
                 if (this.loader) { this.loader.hide(); }
                 this.mainStore.state.setChartIsReady(true);
+                this.mainStore.state.setChartGranularity(getIntervalInSeconds(ViewStore.views[idx].layout));
```
morteza-binary: @shayan-binary it'd be to break the line or assign the granularity in seconds to a const variable and pass it to `setChartGranularity`

PR #893, opened 2019-11-11
shayan/A new modal dialog to show toolbox details on it.

shayan-binary-pr:


src/store/SettingsDialogStore.js
```diff
@@ -7,18 +7,15 @@ export default class SettingsDialogStore {
...
     @observable title = '';
     @observable description = '';
-    @observable id;
 
     @observable activeTab = 'settings'; // 'settings' | 'description'
     @computed get showTabs() { return !!this.description; }
 
     constructor({
-        mainStore, getContext, onDeleted, favoritesId, onChanged,
+        mainStore, getContext, onChanged,
```
morteza-binary: @shayan-binary IMO, we should keep the delete button in the setting dialog, since the default way of opening the setting page is by right clicking on the drown object. I'll talk about it with UI/UX team.

src/components/Dialog.jsx
```diff
@@ -1,18 +1,60 @@
...
+                            </div>
+                        </div>
+                    ) : ''}
+                    <div className="cq-dialog__body">
+                        {children}
+                    </div>
+                </div>
+            </div>
+        ) : (
+            <div
```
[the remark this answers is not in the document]
shayan-binary-pr: @morteza-binary I just add this code, because, we have both menu and dialog, after changing, after we implement all menu to the dialog, we can safely remove this part

src/components/Form.jsx
```diff
@@ -19,7 +19,7 @@ export class Slider extends React.Component {
             step = 1,
             value,
         } = this.props;
-        const barWidth = 120;// css hardcode
+        const barWidth = 238;// css hardcode
```
morteza-binary: @shayan-binary-pr why is it hardcoded?

shayan-binary-pr: @morteza-binary actually it has been hardcoded from the beginning, and because of it's calculation. anyway, I can check to use another slider, because as you mentioned, its not good to be hardcode it

PR #900, opened 2019-11-20
Morteza/proposal_response

morteza-binary:
Added ability to support the second alternative source to draw ticks on the chart.


src/components/ui/Animation.js
```diff
@@ -74,6 +74,7 @@ export default function animateChart(stx, animationParameters) {
     });
 
     stx.append('updateChartData', function (appendQuotes, chart, params) {
```
reza-fs: Is object destruction a norm in the project, I mean, is it possible to use `{ noCreateDataSet }` instead of params in the function signature?

morteza-binary: @reza-fs Yes, we can use; there isn't not any rule for it.

PR #905, opened 2019-12-05
Bala|Kelly/Multiplier

balakrishna-binary:
**Addition to barrier api** 
- hideBarrierLine
    hide/show the barrier line. Can be used to show only the title. Defaults to `false`.
- hideOffscreenLine
    hide/show the barrier line when it is offscreen. Defaults to `false`.
 - hideOffscreenBarrier
    hide/show the barrier line & title when it is offscreen. Defaults to `false`.
 - title
    Title text of the barrier
 - isSingleBarrier
    Shows only High barrier, stops low barrier & shades from rendering when the flag is true. Defaults to `false`.
 - showOffscreenArrows
    hide/show arrows with direction when the barrier is offscreen. Defaults to `false`.
 - opacityOnOverlap
    Sets the opacity of the barrier when it is overlapping with other barrier.

overlap


src/store/PriceLineStore.js
```diff
@@ -60,6 +65,10 @@ export default class PriceLineStore {
         return display;
     }
 
+    @computed get yAxiswidth() {
```
morteza-binary: @balakrishna-binary While `yAxiswidth` is not observable, there is not any need to use `@computed` decorator.

balakrishna-binary: Got it. Thanks!

PR #916, opened 2020-01-04
shayan/Corsshair & zoom

shayan-binary-pr:


src/store/CrosshairStore.js
```diff
@@ -69,15 +69,22 @@ class CrosshairStore {
...
+        this.onChange(state);
         this.mainStore.state.saveLayout();
     }
 
-    setCrosshairState(state) {
+    @action.bound updateProps(onChange) {
+        this.onCrosshairChanged = onChange;
+    }
+
+    onChange(state) {
```
morteza-binary: @shayan-binary-pr `setCrosshairState` would be a better name.

src/store/ChartState.js
```diff
@@ -163,6 +165,20 @@ class ChartState {
             this.comparisonStore.removeAll();
         }
 
+        if (crosshairState !== null && crosshairState !== this.crosshairState) {
+            this.mainStore.crosshair.onChange(crosshairState);
+            this.mainStore.state.saveLayout();
```
morteza-binary: @shayan-binary-pr this line and the line below both are duplicated while there are in `onChange` function.

src/store/ChartState.js
```diff
@@ -72,6 +73,8 @@ class ChartState {
         showLastDigitStats = false,
         startEpoch,
         symbol,
+        crosshairState,
+        chartSize,
```
morteza-binary: @shayan-binary-pr The top widget also need to zoom the chart by setting the number of the bars(ticks) on the screen. since it doesn't know how much should it zoom to get the appropriate zoom level by invoking zoom in/out functions, it would be better to give the ability to set the number visible bars on the screen.

shayan-binary-pr: as we talked, we are going to add new pop for setting `MaxTick`

PR #918, opened 2020-01-07
shayan/set option for download new design 

shayan-binary-pr:


shayan-binary-2:
@morteza-binary  please close this. we did this on another card.

PR #925, opened 2020-02-10
Markers are relocated while scrolling the chart

shayan-binary-2:


src/components/RawMarker.jsx
```diff
@@ -124,8 +125,15 @@ class RawMarker extends React.Component {
             });
             const prices = price_array
                 .map(price => stx.pixelFromPrice(price * 1, chart.panel));
+
+            const canvas = stx.chart.context.canvas;
+            if ((`${canvas.height}`) !== canvas.style.height.replace('px', '')) {
```
morteza-binary: @shayan-binary 
```suggestion
            if (canvas.style.height.indexOf(canvas.height) < 0) {
```

PR #933, opened 2020-02-27
Shayan/Chart Market Selector issue

shayan-binary-2:


src/store/CategoricalDisplayStore.js
```diff
@@ -214,9 +214,12 @@ export default class CategoricalDisplayStore {
 
     @computed get filteredItems() {
         let filteredItems = cloneCategories(this.getCategoricalItems());
+        const activeItmes = this.activeCategories.length
+            ? this.activeCategories
+            : [(this.getCurrentActiveCategory ? this.getCurrentActiveCategory() : 'favorite')];
```
shayan-binary-2: Because we set the render of component for when a user opens it, so `this.activeCategories` not initial unless the component gets opens. and `filteredItems` render after component load.

PR #938, opened 2020-03-04
shayan/Redesign drawing tools dialog

shayan-binary-2:


shayan-binary-2:
we continue this on gitlab

PR #940, opened 2020-03-06
Bala/Assests dialog height fix

balakrishna-binary:
### Changelog
- Style changes for assets dialog


sass/components/_menu-dropdown.scss
```diff
@@ -34,9 +34,15 @@
                     &.cq-menu-dropdown-enter-done {
                         z-index: 6;
                     }
-                    .cq-categorical-display, .cq-dialog__body {
```
shayan-binary-2: @balakrishna-binary what's the issue of changing this line?

balakrishna-binary: `.cq-categorical-display` should not have height as 100% because it has a sibling `title div` which has height `40px`

PR #941, opened 2020-03-08
Shayan/Redesign Indicators dialog

shayan-binary-2:


shayan-binary-2:
close this PR as we use this in another card

PR #944, opened 2020-03-09
zaki/add_more_prefix

zaki-binary:
Changelog
- Add prefixes for `.tooltip` previously missed in last PR.


src/components/Menu.jsx
```diff
@@ -79,7 +79,7 @@ class Menu extends Component {
                         >
                             {first}
                         </div>
-                        {tooltip ? (<div className="tooltip tooltip--right">{tooltip}</div>) : ''}
+                        {tooltip ? (<div className="sc-tooltip sc-tooltip--right">{tooltip}</div>) : ''}
```
shayan-binary-2: @zaki-binary if we suppose to change the class of `tooltip` to something unique related to the Smartchart we should add prefix `ciq`, so the full name would be `ciq-tooltip`. also, we should modify the name in the `main.scss` file.

zaki-binary: My understanding was `ciq` was a prefix for `chartiq`, but we can change to `ciq` if we can confirm there are no existing `ciq-button` or `ciq-tooltip` classNames that might conflict with the general components.

shayan-binary-2: @zaki-binary  you are right `ciq` is a prefix of `chartiq` and yes I check the `main.scss`
so everything is fine. thanks for the explanation .

shayan-binary-2:
@morteza-binary would you please merge this PR

PR #959, opened 2020-04-01
shayan/highest spot not exported to png

shayan-binary-2:


shayan-binary-2:
@zaki-binary  we already fix this issue in another card, as I check the dev branch, it works fine.

PR #969, opened 2020-04-03
update master CI confg

Mahboobeh-binary:


.circleci/config.yml
```diff
@@ -143,41 +120,60 @@ commands:
...
 jobs:
-  build_push_dev:
+  test:
     docker:
       - image: circleci/node:12.13.0-stretch
     steps:
       - git_checkout_from_cache
       - yarn_install
       - test
       - build
```
4p00rv, line 147: we can skip this build step

.circleci/config.yml
```diff
@@ -186,32 +182,46 @@ jobs:
...
     jobs:
-      - build_push_dev:
+      - test:
           filters:
             branches:
-              only: /^dev$/
-          context: binary-frontend-artifact-upload
+              only: 
+                - /^dev/
+                - /pull\/.*/
```
4p00rv, line 192: we can change this to allow all the branches. i.e; it needs to be triggered for merges, PRs, and releases

PR #976, opened 2020-04-08
Shayan/New Design Modifications

shayan-binary-2:


src/store/ChartStore.js
```diff
@@ -146,7 +146,9 @@ class ChartStore {
         const _self = this;
 
         // Add custom injections to the CIQ
-        inject();
+        inject({
+            drawToolsStore: this.mainStore.drawTools,
```
shayan-binary-2, line 150: @morteza-binary  Designer asked us to update the drawing tool hover tooltip with the new naming design. and this is the only way we can update them.

src/store/DrawToolsStore.js
```diff
@@ -83,21 +85,53 @@ export default class DrawToolsStore {
...
 
+    findComputedDrawing = (drawing) => {
+        const group = this.activeToolsGroup.find(drawGroup => drawGroup.key === drawing.name);
+        if (group) {
+            const drawingItem = group.items.find(item => (
+                item.v0 === drawing.v0
+                    && item.v1 === drawing.v1
+                    && item.d0 === drawing.d0
+                    && item.d1 === drawing.d1
+            ));
```
shayan-binary-2, line 119: @morteza-binary its so so rarely that two drawings has the same coordinate, and also we cannot update the chart active drawing store to manage the grouping and new naming, as a result this come to my mind.

PR #977, opened 2020-04-12
Shayan/Portal for setting modals

shayan-binary-2:


app/test.scss
```diff
@@ -18,8 +18,23 @@
...
+            width: 100%;
+            box-sizing: content-box;
+        }
+        input[type='checkbox'] {
+            width: auto;
+        }
+        button {
+            border-radius: 4px;
+            background: #dedede;
+            border: 1px solid #aaa;
```
mahdiaryayi-fs, line 35: Is it OK to use hex colors directly instead of Sass variables?

app/test.scss
```diff
@@ -61,7 +76,12 @@
...
-.portal-node {
-    position: relative;
+.portal-node .cq-modal__overlay {
+    position: fixed;
     z-index: 9;
+    left: 0px;
+    top: 0px;
+    width: 100%;
+    height: 100%;
+    background: rgba(10, 10, 10, 0.3);
```
mahdiaryayi-fs, line 86: same here

src/components/StudyLegend.jsx
```diff
@@ -188,57 +188,61 @@ const StudyLegend = ({
...
+                        </p>
+                        <button
+                            type="button"
+                            className="sc-btn sc-btn--primary sc-btn--w100"
+                            onClick={() => onSelectItem(infoItem.id)}
+                        >
+                            {t.translate('Add')}
+                        </button>
+                    </div>
+                ) : ''}
```
mahdiaryayi-fs, line 229: this ternary's `else` condition in an empty string. so can you use this `{infoItem && <div> ... </div>}` pattern instead and remove the ternary?

app/test.scss
shayan-binary-2: @mahdiaryayi-fs actually this is test file which we use it for developer test and no more. we use to break rules in this file from a long time ago ;)

src/components/StudyLegend.jsx
shayan-binary-2: @mahdiaryayi-fs agree I'll fix this issue

app/test.scss
shayan-binary-2: @mahdiaryayi-fs just like above

PR #982, opened 2020-04-26
Shayan/put synthetic indices on top of markets

shayan-binary-2:


src/binaryapi/ActiveSymbols.js
```diff
@@ -78,9 +79,9 @@ export default class ActiveSymbols {
 
         // Categorize symbols in order defined by another array; there's probably a more
         // efficient algo for this, but for just ~100 items it's not worth the effort
-        const order = ['forex', 'indices', 'stocks', 'commodities', 'synthetic_index'];
+        // const order = ['forex', 'indices', 'stocks', 'commodities', 'synthetic_index'];
```
balakrishna-binary, line 82: can we remove this line?

shayan-binary-2: @balakrishna-binary  that's my bad. just correct it.

PR #983, opened 2020-04-29
Shayan/Fixing Form Dropdown and color

shayan-binary-2:


shayan-binary-2:
this is too old, perhaps we handled it in another card

PR #988, opened 2020-05-14
Shayan/Fixing Form Dropdown and ColorPicker

shayan-binary-2:


sass/components/form.scss
```diff
@@ -279,27 +285,27 @@
...
-        width: 100%;
+        z-index: 2;
         max-height: 200px;
         overflow: scroll;
         box-shadow: 0 2px 8px 0 rgba(42, 48, 82, 0.14);
         box-sizing: border-box;
         border-radius: 4px;
         font-size: 14px;
+        width: 272px;
+        transition: opacity 0.2s, top 0.2s;
```
zaki-binary, line 300: Same here, can we use `transform translate` instead?

sass/components/form.scss
```diff
@@ -439,24 +457,24 @@
...
         right: -5px;
         opacity: 0;
         pointer-events: none;
-        z-index: +2;
+        z-index: 2;
         width: 266px;
         padding: 8px;
-        transition: opacity 0.1s;
         border-radius: 4px;
+        transition: opacity 0.2s, top 0.2s;
```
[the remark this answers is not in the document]
shayan-binary-2: replaced

sass/components/form.scss
shayan-binary-2: replaced

src/store/SettingsDialogStore.js
```diff
@@ -21,13 +22,39 @@ export default class SettingsDialogStore {
...
+            this.setFreezeScroll(false);
+        });
     }
 
     get context() { return this.mainStore.chart.context; }
     get stx() { return this.context.stx; }
     get theme() { return this.mainStore.chartSetting.theme; }
 
     @computed get open() { return this.menu.open; }
+    @action.bound checkDropdownOpen() {
```
balakrishna-binary, line 45: I don't think we need `@action.bound` decorator here as we don't modify any observables

shayan-binary-2: @balakrishna-binary yes you are right. just modified and updated the test link

PR #994, opened 2020-05-18
Shayan/Indicators height issues

shayan-binary-2:


src/store/ChartStore.js
```diff
@@ -132,15 +132,12 @@ class ChartStore {
...
+        const indicatorsHeight = (chartHeight - 320) / 5;
 
         return {
-            MaxAreaHeight: chartHeight - 320, // 320 is the fix height of market selector + floating widget
-            MinHeight: Math.floor(indicatorsMinHeight / 5),
-            MaxHeight: Math.floor(indicatorsMaxHeight / 5),
-            MinPercent: Math.floor(indicatorsMinHeight / 5) / chartHeight,
-            MaxPercent: Math.floor(indicatorsMaxHeight / 5) / chartHeight,
+            height: indicatorsHeight,
+            heightOnAdd: indicatorsHeight + 25,
```
shayan-binary-2, line 139: The CharIQ is not specified exactly heigh we set for indicators, most the time, the height would be the maximum amount. so just for adding, we increase the average height so that make space

PR #996, opened 2020-05-19
Shayan/Form field, Scroll and indicator tooltip issues

shayan-binary-2:


shayan-binary-2:
separate it in different cards, so close this card

PR #998, opened 2020-05-20
Shayan/Update Smartchart packages

shayan-binary-2:


src/components/ChartTypes.jsx
```diff
@@ -5,6 +5,8 @@ import { SettingIcon } from './Icons.jsx';
 import Tooltip from './Tooltip.jsx';
 import '../../sass/components/chart-types.scss';
 
+const TypeIcon = ({ Icon, props }) => (<Icon {...props} />);
```
balakrishna-binary, line 8: Why do we need this wrapper? Is it for better naming?

shayan-binary-2: @balakrishna-binary  just for beauty :sweat_smile:  I just fixed it

PR #1007, opened 2020-06-02
Shayan/remove mobx-react

shayan-binary-2:


shayan-binary-2:
we use another PR https://github.com/binary-com/SmartCharts/pull/1004

PR #1017, opened 2020-06-08
push build artifacts to staging and production branches

Mahboobeh-binary:


.circleci/config.yml
```diff
@@ -75,8 +75,8 @@ commands:
           name: Install and configure dependencies
           command: |
             sudo npm install -g gh-pages@2.0.1
-            git config user.email "ci-build@binary.com"
-            git config user.name "ci-build"
+            git config user.email "mahboub.meng@gmail.com"
+            git config user.name "Mahboobeh-binary"
```
shudarshon-binary, line 79: is there any way to use a machine user github email account in deployment process? that would look good instead of using personal email account

.circleci/config.yml
```diff
@@ -60,36 +59,31 @@ commands:
...
+          name: Tag build
+          command: echo "<< parameters.target_branch >> $(date)" > www/version
+      - run:
+          name: Install and configure dependencies
           command: |
-            tar -cvf artifact.tar index.html nojs-smartcharts.css dist/*
+            sudo npm install -g gh-pages@2.0.1
+            git config user.email "mahboub.meng@gmail.com"
+            git config user.name "Mahboobeh-binary"
+      - add_ssh_keys:
```
shudarshon-binary, line 80: since one git user account is being used to push changes to `gh-pages` branch hence i would suggest the workflow in the following way -> add-ssh-fingerprint-of-machine-user, git-config-user-email-and-name, verify-if-user-is-able-to-ssh-to-github.com (`ssh -Tv git@github.com`), tag-build, deploy-to-branch

.circleci/config.yml
```diff
@@ -60,36 +59,31 @@ commands:
...
-          name: "Compress"
+          name: Tag build
+          command: echo "<< parameters.target_branch >> $(date)" > www/version
+      - run:
+          name: Install and configure dependencies
           command: |
-            tar -cvf artifact.tar index.html nojs-smartcharts.css dist/*
+            sudo npm install -g gh-pages@2.0.1
+            git config user.email "mahboub.meng@gmail.com"
+            git config user.name "Mahboobeh-binary"
```
afshinpaydar-binary, line 79: Is it necessary to hardcode your `git user` and `email`?

.circleci/config.yml
```diff
@@ -145,85 +136,61 @@ jobs:
...
-          context: binary-frontend-artifact-upload
+      - release_aws_production:
           filters:
             branches:
               ignore: /.*/
             tags:
-              only: /^production.*/
-          requires:
-            - build_production
+              only: /^awsproduction.*/
```
afshinpaydar-binary, line 194: `awsproduction.*` only? I didn't see any `awsproduction` tag in releases.

.circleci/config.yml
shudarshon-binary: even i would suggest using git machine user's access control checking to github.com even before the build starts

.circleci/config.yml
Mahboobeh-binary: @shudarshon-binary yup. it was just used for test. will update it.

.circleci/config.yml
Mahboobeh-binary: @shudarshon-binary yes that how we use it, we use machine user

.circleci/config.yml
Mahboobeh-binary: @afshinpaydar-binary we use machine user. it was for testing

PR #1025, opened 2020-06-15
Shayan/ replace tt-react-custom-scrollbars with  native scrollbar

shayan-binary-2:


zaki-binary:
@shayan-binary-2 , if `tt-react-custom-scrollbars` was uninstalled from the devDependencies then how come there are no changes to your `yarn.lock`?

shayan-binary-2:
@zaki-binary  actually it already has changed the `yarn.lock` if you check the bottom of changes, it shows that `tt-react-custom-scrollbars` removed from the `yarn.lock`

shayan-binary-2:
As we decide to use native scrollbar this PR will not use anymore

PR #1044, opened 2020-06-24
Checking CI with NPM

shayan-binary-2:


shayan-binary-2:
close as it was a dummy app

PR #1057, opened 2020-07-05
salar / market close styles

salar-fs:
- Remove gray theme when a market is closed
- Define new color set for when market is closed
- Apply new theme (opacity reduced to 32%) to the charts when market is closed


sass/_themes.scss
```diff
@@ -106,6 +106,8 @@ $COLOR_CORAL_RED  : #ff444f;
 $COLOR_SKY_BLUE   : #2196f3;
 $COLOR_WHITE      : #ffffff;
 $COLOR_BLUE       : #1c5ae3;
+$CHART_BASELINE_DOWN: rgba(249, 84, 84, 0.32);
+$CHART_BASELINE_UP: rgba(76, 175, 80, 0.32);
```
shayan-binary-2, line 110: As mentioned in the design, we should use `0.32` opacity version of original colors, so why need to add new colors? we already have chart types original colors

sass/_themes.scss
```diff
@@ -810,22 +814,26 @@ $themes: (
...
 :export {
     red: $color-red;
+    chart_baseline_up: $CHART_BASELINE_UP;
+    chart_baseline_down: $CHART_BASELINE_DOWN;
     light_chart_text: map-get(map-get($themes, 'light'), 'ChartText');
     dark_chart_text: map-get(map-get($themes, 'dark'), 'ChartText');
     light_chart_grid: map-get(map-get($themes, 'light'), 'ChartGrid');
     dark_chart_grid: map-get(map-get($themes, 'dark'), 'ChartGrid');
+    light_chart_default_border_closed: map-get(map-get($themes, 'light'), 'DefaultBorderClosed');
+    dark_chart_default_border_closed: map-get(map-get($themes, 'dark'), 'DefaultBorderClosed');
```
shayan-binary-2, line 824: Please check with Deveraj, beside of `Area` type the rest of the chart types seems to follow the same colors, green for up and red for down, as a result, we can have a more general definition for them.

shayan-binary-2: Also `DefaultBorder` is the color of the chart grid. better to use another naming for it if we suppose to add a new color for the chart types

sass/_themes.scss
```diff
@@ -810,22 +814,26 @@ $themes: (
 
 :export {
     red: $color-red;
+    chart_baseline_up: $CHART_BASELINE_UP;
+    chart_baseline_down: $CHART_BASELINE_DOWN;
```
shayan-binary-2, line 818: As this is supposed to be the name of the closed graph, I think it would better to rename it to a more meaningful name. Furthermore, If we are going to determine the colors of `Area` chart, its related name is `ChartMountainBg` and `ChartMountainBorder`, and if we are supposed to add a disable color for other types of graph, I think we should use their colors.

src/store/ChartState.js
```diff
@@ -259,51 +259,50 @@ class ChartState {
...
         }
         this.rootNode.querySelector('.chartContainer').style.backgroundColor = Theme[`${theme}_chart_bg`];
-        // change chart colors to grey if the current market is closed and it is not a static chart
+        // reduce opacity to 32% if the current market is closed and it is not a static chart
         if (isChartClosed && !this.isStaticChart) {
-            const closedChartColor = 'rgba(129, 133, 152, 0.35)';
-            this.stxx.setStyle('stx_mountain_chart', 'borderTopColor', closedChartColor);
+            this.stxx.setStyle('stx_mountain_chart', 'borderTopColor', Theme[`${theme}_chart_mountain_border_closed`]);
             this.stxx.setStyle('stx_mountain_chart', 'backgroundColor', Theme[`${theme}_chart_mountain_bg`]);
             this.stxx.setStyle('stx_mountain_chart', 'color', Theme[`${theme}_chart_mountain_bg_shade`]);
```
shayan-binary-2, line 266: Also, please manage with Dervaj, to see, if we should apply opacity on `Area` shade area, its not so obvious in light theme, but dark theme

src/store/ChartState.js
```diff
@@ -259,51 +259,50 @@ class ChartState {
...
+            this.stxx.setStyle('stx_line_up', 'color', Theme[`${theme}_chart_default_border_closed`]);
+            this.stxx.setStyle('stx_line_down', 'color', Theme[`${theme}_chart_default_border_closed`]);
+            this.stxx.setStyle('stx_line_even', 'color', Theme[`${theme}_chart_default_border_closed`]);
             // bar chart
-            this.stxx.setStyle('stx_bar_up', 'color', closedChartColor);
-            this.stxx.setStyle('stx_bar_down', 'color', closedChartColor);
-            this.stxx.setStyle('stx_bar_even', 'color', closedChartColor);
+            this.stxx.setStyle('stx_bar_up', 'color', Theme[`${theme}_chart_default_border_closed`]);
+            this.stxx.setStyle('stx_bar_down', 'color', Theme[`${theme}_chart_default_border_closed`]);
+            this.stxx.setStyle('stx_bar_even', 'color', Theme[`${theme}_chart_default_border_closed`]);
```
shayan-binary-2, line 276: these props are for `OHLC` type, and it has a green/red color. for now, it shows a gray color in close mode.

sass/_themes.scss
salar-fs: These 2 are not in theme colors. They're defined at the very first of `chart.scss` and are used as base colors in the chart.
Right?

sass/_themes.scss
salar-fs: For the first part OK I will talk with him and share the result.
For the second part I agree with you.

src/store/ChartState.js
salar-fs: Will check

sass/_themes.scss
shayan-binary-2: well, for candle type we have the colors we already know, for line chart we have `ChartMountainBg` and `ChartMountainBorder` prop in our theme palate, so as UI/UX defines that colors should be same but with `0.32` opacity for close. so better to change them if line color changed, and for closing, just convert them to `0.32` opacity. so no more new variable for them. am I right?

shayan-binary-2: also for candle, we have `ChartClosedCandleColor` in theme palate

salar-fs:
Closed in favor of #1096

PR #1058, opened 2020-07-07
Shayan/Adding Indicator issues

shayan-binary-2:
We fix three issues in this card

- When adding some special indicators, count down hide
- Y-Axis truncated when adding an indicator
- Indicator description bad style


shayan-binary-2:

PR #1062, opened 2020-07-12
Shayan/Export PNG and Market dropdown on close issues

shayan-binary-2:
- On export PNG navigation widget box appear, which it shouldn't be
 - In mobile mode, when a market is closed the purchase close box overlap the market dropdown


shayan-binary-2:

PR #1064, opened 2020-07-14
Shayan/Adding Indicator issues

shayan-binary-2:
- 1. When adding some special indicators, count down hide
  - 2. Y-Axis truncated when adding an indicator
  - 3. In template duplicate form we need a space between template name


shayan-binary-2:

PR #1082, opened 2020-07-30
Shayan/Current Spot issues

shayan-binary-2:


sass/components/_chart-types.scss
```diff
@@ -30,10 +31,88 @@
...
+.sc-chart-type {
+    display: flex;
+    padding: 24px;
+    justify-content: space-between;
+    @include themify($themes) {
+        background: themed('ChartTypeBg');
+    }
+
+    &__item {
+        $self: &;
```
balakrishna-binary, line 43: Looks like $self is not used anywhere. Can we remove this?

shayan-binary-2: yeah, seems I removed the related code in the refactoring but forget to fix this. thanks.

zaki-binary:
@shayan-binary-2 , can you resolve the conflicts?

PR #1094, opened 2020-08-28
Shayan/update translation for crowdin

shayan-binary-2:


shayan-binary-2:
it's outdate and we have the changes in the master

PR #1095, opened 2020-08-30
Shayan/Market selector issues

shayan-binary-2:
- Fix the issue of the market selector button
 - Clearing search after picking new market

### DIFF
--- a/sass/styles/chart.scss
+++ b/sass/styles/chart.scss
@@ -221,7 +221,7 @@ html, body {
     padding: 0px;
     margin-bottom: 17px;
     border-radius: 4px;
-    z-index: 12;
+    z-index: 15;
     cursor: pointer;
     @include themify($themes) {
         background: themed('NavigationWidgetBg');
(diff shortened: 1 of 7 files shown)


sass/styles/chart.scss
```diff
@@ -221,7 +221,7 @@ html, body {
     padding: 0px;
     margin-bottom: 17px;
     border-radius: 4px;
-    z-index: 12;
+    z-index: 15;
```
salar-fs, line 224: `15` is the minimum value that works? Could we use a variable with a good name for this?

shayan-binary-2: @salar-fs  to be honest, we don't have exact valid range for that. It comes to `15` as 12 seems to have a conflict with barrier shade. that would be a good idea to have constant numbers for them, we should handle it in refactoring

src/store/ChartTitleStore.js
```diff
@@ -51,6 +51,7 @@ export default class ChartTitleStore {
     @observable todayChange = null;
     @observable todayChangePercent = null;
     @observable isVisible = false;
+    @observable enableShowPrice = false;
```
balakrishna-binary, line 54: making `enableShowPrice` as observable is not required here. There is no reaction depending on this. I think we can just use it as normal variable

shayan-binary-2: yes, you are right. nice catch, just updated it

balakrishna-binary (approved):

salar-fs (approved):

PR #1110, opened 2020-10-09
Shayan/Barrier y-axis label change order while dragging

shayan-binary-2:
- Fix overlapping for a barrier for multiplier
- Fix current spot label shape

### DIFF
--- a/sass/components/_barrier.scss
+++ b/sass/components/_barrier.scss
@@ -1,5 +1,9 @@
 @charset "UTF-8";
 
+.barrier {
+    height: 0px;
+}
+
 .barrier-area {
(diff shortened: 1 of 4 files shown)


sass/components/_barrier.scss
```diff
@@ -1,5 +1,9 @@
 @charset "UTF-8";
 
+.barrier {
+    height: 0px;
```
zaki-binary, line 4:
```suggestion
    height: 0;
```

shayan-binary-2: @zaki-binary precise view. just updated

balakrishna-binary (dismissed):

PR #1111, opened 2020-10-15
Shayan/update Circle CI to use package-lock beside yarn.lock

shayan-binary-2:


.circleci/config.yml
```diff
@@ -36,14 +36,14 @@ commands:
       - restore_cache:
           name: Restoring node_modules
           keys:
-            - source-v3-modules-{{ checksum "yarn.lock" }}
+            - source-v3-modules-{{ checksum "package-lock.json" }}
             - source-v3-modules
```
shudarshon-binary, line 40: since CircleCI only stores cache for 15 days so disable this part unless saved with the same key. With first PR you save cache. Then in next PR you enable this part which should be able to restore cache

shudarshon-binary (changes requested):

.circleci/config.yml
```diff
@@ -30,35 +30,35 @@ commands:
...
     steps:
-      - restore_cache:
-          name: Restoring node_modules
-          keys:
-            - source-v3-modules-{{ checksum "yarn.lock" }}
-            - source-v3-modules
+      # - restore_cache:
+      #     name: Restoring node_modules
+      #     keys:
+      #       - source-v3-modules-{{ checksum "package-lock.json" }} CircleCI retention cache for 15 days
```
shudarshon-binary, line 39: make sure you remove "CircleCI retention cache for 15 days" in future. that's just a explanation

shayan-binary-2: @shudarshon-binary, for sure. I'll remove the text in next PR

.circleci/config.yml
shayan-binary-2: thank you for letting me know, as we discussed, let comment the `restore_cache` part in this PR, then we create another PR to return it back, but with new `package-lock.json` name

zaki-binary (approved):

PR #1115, opened 2020-10-25
Shayan/adding feature of active market to chart title

shayan-binary-2:


app/test.jsx
```diff
@@ -452,10 +453,23 @@ class App extends Component {
         window.location.href = `${baseUrl}?marketsOrder=${evt.target.value}`;
     };
 
-    onActiveCategory = () => {
-        this.setState(prevState => ({
-            activeCategory: prevState.activeCategory ? null : 'synthetic_index',
-        }));
+    onActiveMarket = (evt) => {
+        const marketAray = evt.target.value.split(',');
```
balakrishna-binary, line 457:
```suggestion
        const marketArray = evt.target.value.split(',');
```
typo

app/test.jsx
```diff
@@ -251,9 +251,10 @@ class App extends Component {
             <ChartTitle
                 onChange={this.symbolChange}
                 active_category={this.state.activeCategory}
```
balakrishna-binary, line 253: should we remove this line since we don't use `activeCategory` from state anymore?

src/components/ChartTitle.jsx
```diff
@@ -9,6 +9,7 @@ class ChartTitle extends Component {
         const { updateProps: prevUpdateProps, ...previousProps } = prevProps;
         if (
             (previousProps.active_category !== props.active_category)
```
balakrishna-binary, line 11: do we still need this check?

app/test.jsx
shayan-binary-2: @balakrishna-binary just fixed it

app/test.jsx
shayan-binary-2: @balakrishna-binary just fixed it

src/components/ChartTitle.jsx
shayan-binary-2: @balakrishna-binary just fixed it

balakrishna-binary (approved):

shayan-binary-2:
Rename the prop and naming, `open_market` would be more meaningfull

zaki-binary (approved):

PR #1122, opened 2020-11-12
Shayan/Feature load optimization

shayan-binary-2:


README.md
```diff
@@ -149,6 +151,16 @@ Props marked with `*` are **mandatory**:
 | enabledNavigationWidget      | Show or hide navigation widget. Defaults to `false`                                                                                   |
 | isHighestLowestMarkerEnabled | Show or hide the highest and lowest tick on the chart. Defaults to `false`.                                                           |
 
+#### InitialData
+
+Initial data property designed to pass prepared data chart in case you don't want to use Feed or you want to make chart boost for initial. It get below properties and all of them are optional, so if you pass data to them, it cause the chart to use that data, and if you pass `null` it cause the chart to use the Feed call and get api.
```
zaki-binary, line 156:
```suggestion
Initial data property designed to pass prepared chart data in case you don't want to wait for Feed data or if you simply want to make the chart render quicker on its initial load. It gets the properties below and all of them are optional, so if you pass the data, the chart will use that data instead, but if you pass `null` instead, the chart will default to the Feed call and get the data from API.
```

zaki-binary (dismissed):

src/store/ChartStore.js
```diff
@@ -899,6 +916,35 @@ class ChartStore {
...
+     * @param {array} masterData array of ticks regards of desire tick
+     */
+    loadChartWithInitalData(symbol, masterData) {
+        if (!masterData) return;
+
+        const layoutData = createObjectFromLocalStorage(`layout-${this.stateStore.chartId}`);
+        if (!layoutData || !layoutData.symbols.length) return;
+        const layout_symbol = layoutData.symbols[0].symbol;
+
+        this.stxx.loadChart(symbol || layout_symbol || '1HZ10V', {
```
zaki-binary, line 937: Why do we need `1HZ10V` here?

shayan-binary-2: @zaki-binary chart needs a `symbol` for load and this is in case the user forgot to pass any symbol. we can also replace this with an error.

zaki-binary: Maybe replacing it with an error is better? So the user is aware.

shayan-binary-2: @zaki-binary fixed

zaki-binary (approved):

zaki-binary (approved):

PR #1124, opened 2020-11-23
Shayan/Translation configuration

shayan-binary-2:


shayan-binary-2:
we handle the issue in another way, so no need for this PR

PR #1127, opened 2020-11-24
Akmal / Added border radius for download modal

akmal-binary:


sass/components/_download.scss
```diff
@@ -8,6 +8,7 @@
 .sc-download {
     padding: 24px;
     display: flex;
+    border-radius: 4px;
```
salar-fs, line 11: Could we have this as a variable like we have it in **deriv-app**? (`$border_radius`)

akmal-binary: Right, let's use it

salar-fs (approved):

salar-fs:
@shayan-binary-2 Please review this, Mr. SmartChart :pray:

shayan-binary-2 (approved):

zaki-binary (approved):

PR #1133, opened 2020-12-11
Shayan / Refactoring Form component

shayan-binary-2:


src/components/Favorite.jsx
```diff
@@ -1,56 +1,48 @@
...
-    }
-}
+    useEffect(() => {
+        if (store && store.onFavoriteUpdate) {
+            store.onFavoriteUpdate(onFavoriteUpdate);
+            setFavorite(store.isFavorite(category, id));
+        }
+    }, [store, category, id, onFavoriteUpdate]);
+    useEffect(() => {
+        const _store = FavoriteStore.getInstance();
```
balakrishna-binary, line 34: can we use `FavoriteStore.getInstance()` as the default value for the hook at L8 instead of this hook
```
const [store, setStore] = useState(FavoriteStore.getInstance());
```

src/components/Favorite.jsx
```diff
@@ -1,56 +1,48 @@
...
-    }
-
-    onFavoriteUpdate = () => {
-        const { category, id } = this.props;
-        const isFavorite = this.isFavorite(category, id);
-        if (isFavorite !== this.state.isFavorite) {
-            this.setState({ isFavorite });
+const Favorite = ({ category, id }) => {
+    const [store, setStore] = useState(null);
+    const [is_favorate, setFavorite] = useState(false);
```
balakrishna-binary, line 9: typo - `is_favorate`

src/components/Form.jsx
```diff
@@ -35,109 +36,92 @@ export const Checkbox = ({ id, label, checked, disabled, onChange }) => (
...
-                    <input type='range' min={min} max={max} step={step} onChange={this.onChange} value={value} />
-                </div>
-                <div className='value'>{value}</div>
+    const handleChange = useCallback(el => onChange(el.currentTarget.value), [onChange]);
+
+    return (
+        <div className='sc-slider'>
+            <div className='sc-slider-range'>
+                <div className='sc-slider-bar' />
+                <div className='sc-slider-active-bar' style={{ width: getActiveWidth }} />
```
balakrishna-binary, line 52: should we call this function `getActiveWidth`?

shayan-binary-2: @balakrishna-binary it's `useMemo`

balakrishna-binary: maybe we should rename it to `activeWidth`?
`getActiveWidth` sound like a function

shayan-binary-2: @balakrishna-binary agreed. just updated it.

app/DateTimePicker/TimePicker.jsx
```diff
@@ -2,6 +2,7 @@
 /* eslint-disable react/no-array-index-key */
 /* eslint-disable react/no-did-update-set-state */
 /* eslint-disable react/no-unused-state */
+/* eslint-disable max-classes-per-file */
```
balakrishna-binary, line 5:
```suggestion
```
we don't need this line. This line is already there at L1.

src/components/Favorite.jsx
```diff
@@ -1,56 +1,44 @@
-import React, { Component } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
```
zaki-binary, line 1: nitpick: Since we're following `deriv-app` code conventions to , can just keep the import as `React` and we use `React.useCallback`, `React.useEffect` and so on instead?

src/components/Menu.jsx
```diff
@@ -1,158 +1,159 @@
...
+                        open={open}
+                        menu_element={oldDropdown}
+                        portalNodeId={portalNodeId}
+                        onClick={onOverlayClick}
+                    />
+                )) ||
+                    oldDropdown}
+            </div>
+        )) || (
+            <div className={`ciq-menu ciq-disabled ${className || ''}`}>
```
zaki-binary, line 149: Can you use `classNames` here instead of `string template`

zaki-binary (approved):

PR #1135, opened 2020-12-15
Bala/refactor chart and containers

balakrishna-binary:


src/components/ChartTitle.jsx
```diff
@@ -1,80 +1,78 @@
...
-            setMenuOpen,
-            searchInputClassName,
-            theme,
-            onMouseEnter,
-            onMouseLeave,
-            isNestedList,
-        } = this.props;
+const ChartTitle = props => {
+    const {
+        chartId,
```
shayan-binary-2, line 8: just per curiosity, why not replace this with `props` on line `const ChartTitle = props => {` ?

src/components/Chart.jsx
```diff
@@ -21,143 +21,129 @@ import { connect } from '../store/Connect';
...
-                                    {!isChartAvailable && (
-                                        <div className='cq-chart-unavailable'>
-                                            {t.translate('Chart data is not available for this symbol.')}
-                                        </div>
-                                    )}
-                                    <BottomWidgetsContainer>
-                                        <BottomWidget bottomWidgets={bottomWidgets} />
-                                    </BottomWidgetsContainer>
+                <div className='cq-context' ref={rootRef}>
+                    <div className={currentPosition}>
```
shayan-binary-2, line 94: @balakrishna-binary can we use `classNames` lib here and remove `currentPosition`?

balakrishna-binary: done 😄

shayan-binary-2 (approved):

zaki-binary (approved):

PR #1141, opened 2020-12-23
Shayan / Refactoring and cleaning up components

shayan-binary-2:


src/components/MarketOpeningTimeCounter.jsx
```diff
@@ -1,35 +1,26 @@
...
+import React, { useEffect, useMemo, useState } from 'react';
 import { displayMilliseconds } from '../utils/index';
 import ServerTime from '../utils/ServerTime';
 
-export class MarketOpeningTimeCounter extends React.Component {
-    constructor(props) {
-        super(props);
-        this.serverTime = ServerTime.getInstance();
+export const MarketOpeningTimeCounter = ({ symbolOpenTime }) => {
+    const [time, setTime] = useState(0);
```
balakrishna-binary, line 6: can we use `React.useState` to be consistent?

balakrishna-binary (dismissed):

src/components/MarketOpeningTimeCounter.jsx
shayan-binary-2: @balakrishna-binary just fixed it

src/components/MarketOpeningTimeCounter.jsx
```diff
@@ -2,34 +2,25 @@ import React from 'react';
...
-    }
-}
+        return output;
+    }, [symbolOpenTime, time]);
+
+    React.useEffect(() => {
+        const serverTime = ServerTime.getInstance();
+        const timer = setInterval(() => setTime(serverTime.getLocalDate().getTime()), 1000);
+        return () => clearInterval(timer);
+    });
```
balakrishna-binary, line 23: we should add an empty dependency here to mimic `componentDidMount`

PR #1146, opened 2021-01-20
Kseniya/TrackJS - Cannot read property 'noDrag' of null

kseniya-deriv:
- add check for undefined for `myPanel`

### DIFF
--- a/chartiq/development/js/standard.js
+++ b/chartiq/development/js/standard.js
@@ -7473,7 +7473,7 @@ CIQ.ChartEngine.prototype.findHighlights = function (isTap, clearOnly) {
 		this.setMeasure();
 	}
 
-	if (highlightedDraggable && !myPanel.noDrag) {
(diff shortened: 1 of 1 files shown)


chartiq/development/js/standard.js
```diff
@@ -7473,7 +7473,7 @@ CIQ.ChartEngine.prototype.findHighlights = function (isTap, clearOnly) {
 		this.setMeasure();
 	}
 
-	if (highlightedDraggable && !myPanel.noDrag) {
+	if (highlightedDraggable && myPanel && !myPanel.noDrag) {
```
mehrzad-fs, line 7476: you can use optional chaining instead of &&
ref:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
	if (highlightedDraggable && !myPanel?.noDrag) {
```

mehrzad-fs (approved):

akmal-binary (approved):

PR #1147, opened 2021-01-21
Kseniya/TrackJS - Cannot read property 'width' of undefined

kseniya-deriv:


chartiq/development/js/chartiq.js
```diff
@@ -14628,7 +14628,7 @@ CIQ.ChartEngine.prototype.adjustBackingStore = function (canvas, context) {
 };
 
 CIQ.ChartEngine.prototype.reconstituteBackingStore = function () {
-	if (!this.useBackingStore) return;
+	if (!this.useBackingStore || !this.backing) return;
```
shayan-binary-2, line 14631: @kseniya-deriv the `chartiq` library comes from a 3rd party library so we cannot change its code directly. we modify the chartiq code using code injection or monkey patching, you can find injections in `src/chartiq_injections/backingStore.js`

shayan-binary-2 (changes requested):
we need to use injection beside changing the code directly

src/chartiq_injections/backingStore.js
```diff
@@ -16,4 +16,21 @@ export const BackingStore = () => {
         this.adjustedDisplayPixelRatio = 1;
         this.draw();
     };
+
+    CIQ.ChartEngine.prototype.reconstituteBackingStore = function () {
+        if (!this.useBackingStore || !this.backing) return;
```
shayan-binary-2, line 21: @kseniya-deriv just to make a more clear, it's better to put a note here that we changed this line.

shayan-binary-2 (approved):

carol-binary (approved):

shayan-binary-2 (approved):

zaki-binary (approved):

PR #1148, opened 2021-01-29
Shayab / Bug chart store and chart issues

shayan-binary-2:


shayan-binary-2:
The issue no more happen, let archive this card

PR #1149, opened 2021-01-29
Shayan / Fix main store initial part

shayan-binary-2:


shayan-binary-2:
we fixed this issue in another card

PR #1155, opened 2021-02-11
Shayan / Upgrade peer dependency

shayan-binary-2:


shayan-binary-2:
We consider another solution, this PR is not useful anymore

PR #1180, opened 2021-06-22
Shayan / modifying open_market and README file

shayan-binary-2:


src/store/ChartTitleStore.js
```diff
@@ -184,9 +184,9 @@ export default class ChartTitleStore {
         this.enableShowPrice = false;
     }
 
-    @action.bound updateProps({ open_market, open }) {
-        if (open_market) {
-            this.openMarket = open_market;
+    @action.bound updateProps({ open_market, openMarket, open }) {
+        if (open_market || openMarket) {
+            this.openMarket = open_market || openMarket;
```
matin-binary, line 189: I don't think there is a need for `openMarket`, we can extend `open_market` and extend it.

matin-binary (approved):

PR #1196, opened 2021-09-03
Maryia/Avoiding error if myPanel is null

maryia-deriv:


chartiq/development/js/standard.js
```diff
@@ -7566,7 +7566,7 @@ CIQ.ChartEngine.prototype.findHighlights = function (isTap, clearOnly) {
 		this.setMeasure();
 	}
 
-	if (highlightedDraggable && !myPanel.noDrag) {
+	if (highlightedDraggable && !myPanel?.noDrag) {
```
balakrishna-binary, line 7569: This condition will be truthy if `myPanel` is undefined too. We should check if `myPanel` is defined first.

Suggested Change:
```
if (highlightedDraggable && myPanel && !myPanel.noDrag) {
```

maryia-deriv: Done, Thank you)

balakrishna-binary (approved):

PR #1199, opened 2021-09-09
Maryia/Avoiding the 'cannot read getTime/getSeconds/getMinutes of undefined' error

maryia-deriv:

PR #1203, opened 2021-09-15
Maryia/Fix for 'Cannot read getTime() of undefined' error

maryia-deriv:


chartiq/development/js/chartiq.js
```diff
@@ -11452,7 +11452,7 @@ CIQ.ChartEngine.prototype.dateFromTick = function (
 	}
 
 	if (nativeDate) {
-		result = new Date(dt.getTime());
+		result = new Date(dt?.getTime());
```
balakrishna-binary, line 11455: can we find why `dt` is undefined? may be we are not passing it?
`new Date(undefined)` will be an Invalid date.

maryia-deriv: I have done it in another PR: https://github.com/binary-com/SmartCharts/pull/1199 😊

PR #1204, opened 2021-09-19
Maryia/Fix for undefined startEpoch & if suggestedStartDate > endEpoch

maryia-deriv:


src/feed/Feed.js
```diff
@@ -144,7 +144,11 @@ class Feed {
         const granularity = calculateGranularity(period, interval);
         const key = this._getKey({ symbol, granularity });
         const localDate = this._serverTime.getLocalDate();
-        suggestedStartDate = suggestedStartDate > localDate ? localDate : suggestedStartDate;
+        if (suggestedStartDate > localDate) {
+            suggestedStartDate = localDate;
+        } else if (!this.startEpoch && suggestedStartDate > this.endEpoch) {
+            suggestedStartDate = CIQ.strToDateTime(getUTCDate(this.endEpoch - 1000));
+        }
```
matin-binary, line 151: Do we need to include `else` in this condition?

maryia-deriv: I agree that we don't 👍🏻  Thank you)

matin-binary (dismissed):

balakrishna-binary (dismissed):

matin-binary (dismissed):

matin-binary (dismissed):

src/feed/Feed.js
```diff
@@ -395,6 +398,7 @@ class Feed {
 
         if (
             this.endEpoch &&
+            this._stx.chart.dataSet.slice(-1)[0] &&
```
balakrishna-binary, line 401: should we check if `this._stx.chart.dataSet` is defined?

maryia-deriv: I can add this check too, however we should also check if **this._stx.chart.dataSet.slice(-1)[0]** is defined because at the time when we get the console error, **this._stx.chart.dataSet.slice(-1)** is defined. Do you want me to add the check for this._stx.chart.dataSet as well despite this fact? :)

balakrishna-binary: Okay, it's fine if `dataSet` is not undefined at any point.

balakrishna-binary (approved):

PR #1206, opened 2021-09-27
Maryia/Avoiding error in case this.p0 or this.p1 is undefined or null

maryia-deriv:


balakrishna-binary (approved):

maryia-deriv:
the fix has been merged into PR #1204

PR #1209, opened 2021-10-17
Mahdiyeh/ Fix: disconnect error in resizeObserver

mahdiyeh-fs:


chartiq/development/js/chartiq.js
```diff
@@ -3141,7 +3141,7 @@ CIQ.resizeObserver = function (element, listener, resizeHandle, timeout) {
 		}
 	} else {
 		if (resizeHandle) {
-			if (typeof ResizeObserver !== "undefined") {
+			if (typeof ResizeObserver !== "undefined" && resizeHandle.disconnect === 'function' ) {
```
matin-binary, line 3144: Do we need to add `typeof` ?

mahdiyeh-fs: you're right, my bad :(

PR #1220, opened 2021-11-09
MaryiaF/feat: market icon align at the right corner on the mobile ver…

maryiafrantsava-deriv:
…sion


akmal-binary (approved):

maryiafrantsava-deriv:
I noticed that the currency icons are slightly different in design in production. When I started checking what was going on, I noticed that, for example, the 'volatility 75 Index' icon is the same as the 'Jump 75 Index'.
Screenshot 2021-11-12 at 14 43 11
It turned out that the problem is in the file names: the file names are the same. Despite the fact that the folders in which they are stored are different. Apparently, these are the features of compiling sprites.
That’s why I renamed the files and it all worked.

akmal-binary (dismissed):

balakrishna-binary (approved):

matin-binary (approved):

akmal-binary (approved):

PR #1229, opened 2021-11-29
Maryia/Fix for console error 'Cannot read properties of undefined (reading 'replace')'

maryia-deriv:


src/store/CrosshairStore.js
```diff
@@ -241,7 +241,7 @@ class CrosshairStore {
                             sKey = CIQ.createObjectChainNames(sKey, subField)[0];
                         }
                         const display = seriesParams.display || seriesParams.symbol || seriesParams.field;
-                        if (sKey && (!dupMap[display] || seriesParams.symbol === undefined)) {
+                        if (display && sKey && (!dupMap[display] || seriesParams.symbol === undefined)) {
```
matin-binary, line 244: this condition does not necessarily need `display` as we have `||` operator here `(!dupMap[display] || seriesParams.symbol === undefined)`.

Do we have to add it to the condition?

matin-binary (approved):

PR #1230, opened 2021-11-29
Maryia/Replaced direct changes to standard.js with injection to avoid error if myPanel is null

maryia-deriv:
Have to revert previous direct changes to standard.js and add chartiq injection instead because previous changes didn't affect the library file.


src/chartiq_injections/findHighlights.js
```diff
@@ -0,0 +1,543 @@
...
+                        somethingChanged = true;
+                    }
+                if (drawing.highlighted) {
+                    stickyArgs.noDelete = drawing.permanent;
+                    stickyArgs.noEdit = !this.callbackListeners.drawingEdit.length;
+                }
+            }
+        }
+
+        let n, o, m, marker, series, bar;
```
maryia-deriv, line 79: Note: had to replace all vars with `let`s & `const`s that's why I declare this `bar` variable here instead of declaring `var bar = ...` at line 307 for the first time.

src/chartiq_injections/findHighlights.js
```diff
@@ -0,0 +1,543 @@
...
+                        if (o.study.isHighlighted(this, cx, cy)) {
+                            o.highlight = true;
+                            this.anyHighlighted = true;
+                        }
+                        continue;
+                    }
+
+                    const quote = chart.dataSegment[bar];
+                    if (!quote) continue;
+                    let out;
```
maryia-deriv, line 328: Note: had to replace all vars with `let`s & `const`s that's why I introduce this `out` variable here instead of declaring `var out` at line 329 for the first time.

src/chartiq_injections/findHighlights.js
```diff
@@ -0,0 +1,543 @@
...
+        if ((drag === true || (drag && drag.yaxis)) && myYAxis && !myYAxis.noDraw) {
+            this.anyHighlight = true;
+            highlightedDraggable = myYAxis;
+        }
+
+        if (!this.anyHighlighted) {
+            this.setMeasure();
+        }
+
+        if (highlightedDraggable && myPanel && !myPanel.noDrag) {
```
maryia-deriv, line 510: Note: This is a fix that I previously added directly to standard.js.

balakrishna-binary (approved):

PR #1233, opened 2021-11-30
Hubert / fixing TrackJS stx null error

hubert-deriv:


src/store/Connect.js
```diff
@@ -37,6 +37,7 @@ function connectCustomStore(mapperFunction, CustomStore) {
             componentWillUnmount() {
                 if (this.store.destructor) {
                     this.store.destructor();
+                    this.store.mainStore.drawTools.destructor();
```
balakrishna-binary, line 40: This will call `drawTools.destructor` on every connected component's unmount

src/store/DrawToolsStore.js
```diff
@@ -144,6 +144,10 @@ export default class DrawToolsStore {
         return null;
     };
```
balakrishna-binary, line 146: I think we can set up a reaction for this

Let's get rid of `when(() => !!this.context, this.onContextReady);`
and replace it with

```
reaction(() => this.context, () => { 
    if (!!this.context) {
        this.onContextReady();
    } else {
        this.destructor();
    }
});
```

hubert-deriv: Aaah I see! Thanks for the suggestion, I added this.

balakrishna-binary (approved):

PR #1239, opened 2021-12-12
Mahdiyeh/Fix: prevent volatility chart from loading continiuesly

mahdiyeh-fs:


src/store/LastDigitStatsStore.js
```diff
@@ -56,7 +56,7 @@ export default class LastDigitStatsStore {
         return this.mainStore.state.shouldMinimiseLastDigits;
     }
 
-    @action.bound async updateLastDigitStats(response = {}) {
+    @action.bound async updateLastDigitStats() {
```
balakrishna-binary, line 59:
```suggestion
    @action.bound async updateLastDigitStats(response) {
```
Removing the default assignment alone should fix the problem.

src/store/LastDigitStatsStore.js
```diff
@@ -71,12 +71,10 @@ export default class LastDigitStatsStore {
...
-            const tickHistory =
-                response ||
-                (await this.api.getTickHistory({
-                    symbol: this.mainStore.chart.currentActiveSymbol.symbol,
-                    count: this.count,
-                }));
+            const tickHistory = await this.api.getTickHistory({
+                symbol: this.mainStore.chart.currentActiveSymbol.symbol,
+                count: this.count,
+            });
```
balakrishna-binary, line 77: This change can be reverted

PR #1242, opened 2021-12-28
Maryia/Updating Favorites instantly & keeping logics for translation without page reload

maryia-deriv:
Fix: Reverting action back to computed value to make Favorites update instantly again


balakrishna-binary (approved):

matin-binary (dismissed):

ali-hosseini-fs:
Since we need both of the patterns, Action and computed values, for the translation and for favorites to work as expected.
Please apply the changes that we agreed to separate the data from the   `favoritesCategory` object and use it from the computed values.

maryia-deriv:
> Since we need both of the patterns, Action and computed values, for the translation and for favorites to work as expected. Please apply the changes that we agreed to separate the data from the `favoritesCategory` object and use it from the computed values.

@ali-hosseini-fs Thank you so much for finding the most suitable option to make our changes combined work well! 😊

ali-hosseini-fs (dismissed):
LGTM, Thanks

matin-binary (dismissed):

PR #1256, opened 2022-01-31
MaryiaF/fix: platform settings text is truncated when the language is…

maryiafrantsava-deriv:
… changed


sass/components/_dialog.scss
```diff
@@ -90,14 +90,15 @@
             border-radius: 0 5px 0 0;
 
             .ic-icon {
-                width: 32px;
-                height: 32px;
-                line-height: 34px;
+                width: 27px;
```
maryiafrantsava-deriv, line 93: I had to change it because the old dimensions added padding from the bottom after removing the fixed block height. After my changes, the cross remained in its place.

sass/components/_dialog.scss
```diff
@@ -86,7 +86,7 @@
             padding: 10px 0px 10px 12px;
         }
         &--action {
-            padding: 16px;
```
mahdiyeh-fs, line 89: suppose it's better to add `{
                                              display: flex;
                                              justify-items: center;
                                              padding: 16px;
                                          }`

sass/components/_dialog.scss
```diff
@@ -98,6 +98,7 @@
                 padding: 4px;
                 position: relative;
                 top: -6px;
+                right: -2.5px;
```
mahdiyeh-fs, line 101: and here removee both `top` and `right` positioning

maryiafrantsava-deriv: @mahdiyeh-fs, Here similar situation: if I remove them, the design will be changed. For a block with class 'sc-dialog__head' will be adding additional indent, which wasn’t there before fixing the bug.

sass/components/_dialog.scss
maryiafrantsava-deriv: @mahdiyeh-fs , without `display: flex; justify-items: center; ` result will be the same. You can check. if I will use `padding: 16px; `instead of `padding: 16px 16px 8px 8px; , then design will be changed. I'm not sure that’s it good practice to change layout where it’s not required.

mahdiyeh-fs: changing the padding to `0 16px;` will fix that.

matin-binary (approved):

balakrishna-binary (approved):

PR #1260, opened 2022-02-17
Akmal / Fix ticks websocket problem

akmal-binary:


src/feed/Feed.ts
```diff
@@ -443,10 +448,18 @@ class Feed {
...
+            // we will reload the chart instead of rendering new ticks
+            // in case if the duration between the previous tick and the current one is longer than a minute
+            // to avoid disturbing animation on slower devices
+            if (current_tick_timestamp - this._last_tick_timestamp < 60) {
+                this._stx.updateChartData(quotes, null, {
+                    allowReplaceOHL: true,
+                });
+                this._stx.createDataSet();
+            } else {
+                this._mainStore.chart.refreshChart();
```
balakrishna-binary, line 460: Will this be called multiple times when multiple ticks come at once?

matin-binary (approved):

src/feed/Feed.ts
akmal-binary: No, it will only be executed once when the difference between the previous and the current tick is above 60 seconds. After that the chart will refresh and reset the whole feed, so it will have at least another 60 seconds before the next execution.

balakrishna-binary (dismissed):

matin-binary (dismissed):

maryia-binary (approved):

balakrishna-binary (approved):

PR #1261, opened 2022-02-18
Maryia/TrackJS_Avoiding_fullscreen_error_unhandled_rejection

maryia-deriv:


src/store/ChartStore.ts
```diff
@@ -1246,9 +1246,11 @@ class ChartStore {
         const fncToCall = fullscreen_map[isInFullScreen ? 'fnc_exit' : 'fnc_enter'].find(
             fnc => (el as HTMLElement)[fnc as keyof HTMLElement]
         );
-
+        // fncToCall can be undefined for iOS' Safari that does not support fullscreenAPI
         if (fncToCall) {
-            (el as HTMLElement)[fncToCall as 'requestFullscreen']();
+            (el as HTMLElement)[fncToCall as 'requestFullscreen']()?.catch(() => {
```
maryia-deriv, line 1251: '?' is needed to avoid 'TypeError: undefined is not an object (evaluating 'el[fncToCall]()["catch"]')' error.

src/store/ChartStore.ts
```diff
@@ -1246,9 +1246,11 @@ class ChartStore {
...
+                /* no actual handling expected */
+            });
```
maryia-deriv, line 1253: '?' before .catch() is needed to avoid 'TypeError: undefined is not an object (evaluating 'el[fncToCall] () ["catch"]')' error.
And .catch() contains an empty callback without handling because there are 2 possible cases when requestFullscreen() can return a rejected promise with 'fullscreen error' reason here:
1. The document containing the element isn't fully active; that is, it's not the current active document.
2. The element is not permitted to use the 'fullscreen' feature, either because of Feature Policy configuration or other access control features.
Please correct me if I'm wrong but I believe that this rejection cannot lead to app crash, and handling a rejection with an empty callback will help avoid this track js error.

matin-binary (approved):

src/store/ChartStore.ts
habib-deriv: As you want to ignore them, you must do so explicitly. You can do it like this,
`somePromise.catch(e => void e);`

maryia-deriv: Thank you so much @habib-deriv ) I applied your suggestion but I also had to put `undefined` instead of `void e` in order to avoid _eslint no-void error_ 🙂

habib-deriv (approved):
LGTM

PR #1270, opened 2022-03-11
Maryia/Fix: Set isOnPagination to false when pagination finished fetching

maryia-deriv:
Fix: Set isOnPagination to false when pagination finished fetching


src/feed/Feed.ts
```diff
@@ -335,7 +337,6 @@ class Feed {
                 if (response.error) {
                     const { message: text } = response.error as TError;
                     this.loader.hide();
-                    this.paginationLoader.updateOnPagination(false);
```
maryia-deriv, line 338: A small note just in case: This line is not needed anymore here because in case response.error === true, it will quit _getPaginationData() upon `return` at line 346 and execute `this.paginationLoader.updateOnPagination(false)` at line 304 instead.

mahdiyeh-fs (approved):

matin-binary (approved):

balakrishna-binary (approved):

PR #1271, opened 2022-03-21
Akmal / Fix: do not close forex markets

akmal-binary:


src/components/categoricaldisplay/Item.tsx
```diff
@@ -31,7 +31,7 @@ const ItemName = React.memo(({ item: { itemId, display } }: { item: TSubCategory
 const ItemDetail = React.memo(
     ({ favoritesId, item: { dataObject, itemId } }: { favoritesId: string; item: TSubCategoryDataItem }) => (
         <div className='sc-mcd__item__detail'>
-            {dataObject && (dataObject.exchange_is_open === undefined || dataObject.exchange_is_open) ? (
+            {dataObject && (dataObject.market === 'forex' || dataObject.exchange_is_open === undefined || dataObject.exchange_is_open) ? (
```
yashim-deriv, line 34:
```suggestion
            {dataObject && (dataObject.market === 'forex' || dataObject.exchange_is_open === undefined || dataObject.exchange_is_open) ? (
```
just a question. are we specifically checking for `undefined`?  looks kinda sus 👀

akmal-binary: Yes 😁 It was added a long time ago as well 😄

akmal-binary: I think the point of this was that it should only show `CLOSED` when the market is explicitly closed, and not when the response is undefined/not received yet

src/components/SymbolSelectButton.tsx
```diff
@@ -62,12 +62,13 @@ export const SymbolInfo = observer(SymbolInfoBase);
...
     const { chartTitle } = useStores();
-    const { isSymbolOpen } = chartTitle;
+    const { isSymbolOpen, currentActiveCategory } = chartTitle;
+    const is_forex_market = currentActiveCategory === 'forex';
     return (
         <div className='cq-symbol-select-btn'>
             <div className='cq-symbol-select-btn__container'>
                 <SymbolInfo />
-                {!isSymbolOpen && <div className='cq-symbol-closed-text'>{t.translate('CLOSED')}</div>}
+                {!isSymbolOpen && !is_forex_market && <div className='cq-symbol-closed-text'>{t.translate('CLOSED')}</div>}
```
balakrishna-binary, line 71: I think sometimes Forex markets can be closed from BO. Will it still show `Closed` here if it is closed in BO?

akmal-binary: Hmm, no, unfortunately this is required due to forex markets being marked as closed when the daily blackout happens. But in case the markets are closed from BO, the user will not be able to make any trade and it will show the error on the right side. I think the idea of the card is that the forex markets do not close in general, so if trading is stopped for some reason, it still should not be marked as closed on our side.

balakrishna-binary (approved):

yashim-deriv (approved):
LGTM

matin-binary (approved):

src/components/SymbolSelectButton.tsx
balakrishna-binary: @akmal-binary Looks like `Forex` market will be closed on Weekends. We still need to show `Closed` label for it.
Screenshot 2022-03-28 at 12 01 36 PM

akmal-binary: Thanks Bala, I'll change the logic accordingly 👌🏻

PR #1272, opened 2022-03-24
MaryiaF/fix: trackJS issue Cannot read properties of undefined (readi…

maryiafrantsava-deriv:
…ng '0')

### DIFF
--- a/src/store/DrawToolsStore.ts
+++ b/src/store/DrawToolsStore.ts
@@ -290,7 +290,7 @@ export default class DrawToolsStore {
             item = drawTools[item.name] ? { ...item, ...drawTools[item.name] } : item;
             item.index = indx;
             item.bars =
-                ignoreBarType.indexOf(item.name) === -1 ? item.p1[0] && item.p0[0] : null ? Math.abs(Math.floor(item.p1[0] - item.p0[0])) + 1 : null;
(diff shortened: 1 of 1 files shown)


src/store/DrawToolsStore.ts
```diff
@@ -290,7 +290,7 @@ export default class DrawToolsStore {
             item = drawTools[item.name] ? { ...item, ...drawTools[item.name] } : item;
             item.index = indx;
             item.bars =
-                ignoreBarType.indexOf(item.name) === -1 ? Math.abs(Math.floor(item.p1[0] - item.p0[0])) + 1 : null;
+                ignoreBarType.indexOf(item.name) === -1 ? item.p1[0] && item.p0[0] : null ? Math.abs(Math.floor(item.p1[0] - item.p0[0])) + 1 : null;
```
balakrishna-binary, line 293: I think there are some cases when p1 or p0 is undefined. That's when it will throw `Cannot read properties of undefined`.

we need to check truthy on `item.p1 && item.p0` instead

maryiafrantsava-deriv: i've changed

matin-binary (approved):

balakrishna-binary (approved):

PR #1291, opened 2022-06-01
Thisyahlen/save template bug fix

thisyahlen-deriv:
Saved templates not loading the proper templates.

### DIFF
--- a/src/store/ViewStore.ts
+++ b/src/store/ViewStore.ts
@@ -93,7 +93,7 @@ export default class ViewStore {
     @action.bound saveViews() {
         if (ViewStore.views.some(x => x.name.toLowerCase().trim() === this.templateName.toLowerCase().trim())) {
             this.updateRoute('overwrite');
-        } else if (this.templateName.toLowerCase().trim().length > 0) {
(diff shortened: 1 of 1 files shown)


src/store/ViewStore.ts
```diff
@@ -135,7 +135,7 @@ export default class ViewStore {
         }
         this.mainStore.state.setChartIsReady(false);
         const stx = this.stx;
-        const granularity = getIntervalInSeconds(ViewStore.views[idx].layout) as TGranularity;
+        const granularity = getIntervalInSeconds(this.sortedItems[idx].layout) as TGranularity;
```
matin-binary, line 138: - You can define a variable for sortedItems and use it multiple places instead of calling the sort function multiple times.
- Why are you passing `idx` to the function?

thisyahlen-deriv: Alrighty will create a variable. 
For the idx, so that the system will know which template im clicking. if not it will return all the idx

matin-binary (dismissed):

balakrishna-binary (approved):

balakrishna-binary (approved):

PR #1336, opened 2023-01-13
feat: yauheni/ remove barriers pip size

yauheni-deriv:
Allow smartcharts barriers to accept and show any values from BE without rounding

### DIFF
--- a/src/components/PriceLine.tsx
+++ b/src/components/PriceLine.tsx
@@ -29,7 +29,6 @@ const PriceLine = ({
     store,
 }: TPriceLineProps) => {
     const {
-        price,
         priceDisplay,
         visible,
         setDragLine,
@@ -74,7 +73,7 @@ const PriceLine = ({
                 )}
                 <div className='draggable-area' />
                 <div className='drag-price' style={{ backgroundColor: color, width, opacity }}>
-                    <div className='price'>{isDragging? priceDisplay : price}</div>
+                    <div className='price'>{priceDisplay}</div>
                     {offScreen && offScreenDirection && (
                         <PriceLineArrow offScreenDirection={offScreenDirection} color={color} />
                     )}
--- a/src/components/PriceLine.tsx
+++ b/src/components/PriceLine.tsx
@@ -29,6 +29,7 @@ const PriceLine = ({
     store,
 }: TPriceLineProps) => {
     const {
+        price,
         priceDisplay,
         visible,
         setDragLine,
@@ -73,7 +74,7 @@ const PriceLine = ({
                 )}
                 <div className='draggable-area' />
                 <div className='drag-price' style={{ backgroundColor: color, width, opacity }}>
-                    <div className='price'>{priceDisplay}</div>
+                    <div className='price'>{isDragging? priceDisplay : price}</div>
                     {offScreen && offScreenDirection && (
                         <PriceLineArrow offScreenDirection={offScreenDirection} color={color} />
                     )}
--- a/src/store/BarrierStore.ts
+++ b/src/store/BarrierStore.ts
@@ -2,7 +2,6 @@ import { action, computed, observable, makeObservable } from 'mobx';
 import MainStore from '.';
 import Context from '../components/ui/Context';
 import { TBarrierChangeParam, TBarrierUpdateProps, TCIQAddEventListener, TCIQAppend } from '../types';
-import { isValidProp } from '../utils';
 import PendingPromise from '../utils/PendingPromise';
 import PriceLineStore from './PriceLineStore';
 import ShadeStore from './ShadeStore';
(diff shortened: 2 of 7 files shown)


src/store/BarrierStore.ts
```diff
@@ -275,8 +271,8 @@ export default class BarrierStore {
     }
 
     _fireOnBarrierChange = (): void => {
-        const high = this._high_barrier.visible ? +this._high_barrier.price.toFixed(this.pip) : undefined;
-        const low = this._low_barrier.visible ? +this._low_barrier.price.toFixed(this.pip) : undefined;
+        const high = this._high_barrier.visible ? +this._high_barrier.price : undefined;
+        const low = this._low_barrier.visible ? +this._low_barrier.price : undefined;
```
balakrishna-deriv, line 275: should we make this change for accumulators?
This is breaking higher/lower contracts when we drag the barrier.

src/components/PriceLine.tsx
```diff
@@ -73,7 +74,7 @@ const PriceLine = ({
                 )}
                 <div className='draggable-area' />
                 <div className='drag-price' style={{ backgroundColor: color, width, opacity }}>
-                    <div className='price'>{priceDisplay}</div>
+                    <div className='price'>{isDragging? priceDisplay : price}</div>
```
yauheni-deriv, line 77: this condition allows to show correct number while dragging the barriers in high/low and touch/ no touch contracts.
because while dragging price became not rounded and looks ugly (f.e: 9899,4634762347624 instead of 9899,463 having a pipsize 3)

src/store/ChartStore.ts
```diff
@@ -1051,10 +1048,10 @@ class ChartStore {
         }
     }
     calculateYaxisWidth = (price: number) => {
-        if (!price) return;
+        if (!price || this.isBarrierDragging) return;
```
yauheni-deriv, line 1052: this flag prevent calculating the width of the yAxis while dragging. Because it can became too wide because of deleting pip size.

matin-deriv (approved):
LGTM

src/components/PriceLine.tsx
akmal-deriv: Hey @yauheni-deriv , is this change requested in your card? I think it make sense to have a more specific value while dragging the barriers to be able to drag it more accurately. Unless the number of decimals is indeed too big.

src/store/ChartStore.ts
```diff
@@ -1051,10 +1048,10 @@ class ChartStore {
...
 
         const { context } = this.context?.stx.chart;
-        const priceWidth = context.measureText(price.toFixed(this.pip)).width + 20;
+        const priceWidth = context.measureText(price.toString()).width + 20;
```
akmal-deriv, line 1054: Have you checked whether this works as expected with different currencies/markets and trade types? I believe depending on these parameters, the amount of decimal points differs, that's why there was a function to fix it.

src/components/PriceLine.tsx
yauheni-deriv: without these changes we couldn’t see real BE values. Because 'priceDisplay' consider pipsize of the symbol. and for other contracts which require barriers to be adjustable the behaviour will remain the same while dragging.
And if there are some issues with BE values for barriers of other contracts BE will fix it as was agreed

src/store/ChartStore.ts
yauheni-deriv: I checked few of others.. I'll ask qa to check more detailed. without fixing it we will be unable to adjust yaxis according to new barriers values

akmal-deriv (approved):

jim-deriv (approved):

balakrishna-deriv (approved):

src/utils/index.ts
```diff
@@ -17,8 +17,8 @@ export function createObjectFromLocalStorage(key: string) {
     return undefined;
 }
 
-export function isValidProp(p: number) {
-    return p !== undefined && !isNaN(p); // eslint-disable-line no-restricted-globals
+export function getStringValue(p: number | string) {
+    return (typeof p === 'string') ? p : p.toString();
```
niloofar-deriv, line 21:
```suggestion
    return (typeof p == 'string') ? p : p.toString();
```

niloofar-deriv (approved):
LGTM :+1:

src/utils/index.ts
yauheni-deriv: thank for the suggestion, but I believe we don’t need here type coercion.. '===' is working fine

maryia-deriv (approved):
right after i approved the pr, i noticed a small issue with the '+' sign shown only when a user is dragging the barrier for example for Vol 100 (1s) Index in Higher/Lower. The sign disappears after a user stops dragging the barrier. Please try reverting < div className='price' >{isDragging? priceDisplay : price}< /div > in src/components/PriceLine.tsx to the old < div className='price' >{priceDisplay}< /div > and modify `priceDisplay` getter in src/store/PriceLineStore.ts so that it returns the correct value.

yauheni-deriv:
> right after i approved the pr, i noticed a small issue with the '+' sign shown only when a user is dragging the barrier for example for Vol 100 (1s) Index in Higher/Lower. The sign disappears after a user stops dragging the barrier. Please try reverting < div className='price' >{isDragging? priceDisplay : price}< /div > in src/components/PriceLine.tsx to the old < div className='price' >{priceDisplay}< /div > and modify `priceDisplay` getter in src/store/PriceLineStore.ts so that it returns the correct value.

the issue had been fixed

maryia-deriv (approved):
👍🏻

akmal-deriv (dismissed):

niloofar-deriv (dismissed):
LGTM :+1:

maryia-deriv (approved):

niloofar-deriv (approved):
LGTM :+1:

balakrishna-deriv (approved):

PR #1344, opened 2023-01-23
kate/86796/remove console warnings from SmartCharts localhost

kate-deriv:
In FilterGroup and ResultsPanel components there was no key after .map method and that caused console errors.
Delete extra 'else' statement, because in 'if' we have return.


maryia-deriv (approved):
👍🏻

mahdiyeh-deriv (approved):

matin-deriv (approved):

PR #1383, opened 2023-03-16
Akmal / fix: unwanted api call when different chart is opened

akmal-deriv:


src/store/ChartStore.ts
```diff
@@ -741,6 +741,7 @@ class ChartStore {
         } = props;
         this.feedCall = feedCall || {};
         this.api = new BinaryAPI(requestAPI, requestSubscribe, requestForget, requestForgetStream);
+        this.currentLanguage = localStorage.getItem('current_chart_lang')?.toLowerCase();
```
balakrishna-deriv, line 744: When we load for the first time, `this.currentLanguage` will be undefined, right?

akmal-deriv: Yes, correct, that's why there will be one API call to set the language for the first time. We can't really do anything, as without this, there's a language bug which will not change language for market switcher

balakrishna-deriv (dismissed):

balakrishna-deriv (dismissed):

balakrishna-deriv (approved):

maryia-deriv (approved):

maryia-deriv (approved):

kate-deriv (approved):

PR #1387, opened 2023-03-29
Akmal / feat: remove relative indicator if it's blocking current price

akmal-deriv:


henry-deriv (approved):

src/components/PriceLine.tsx
```diff
@@ -73,7 +76,7 @@ const PriceLine = ({
                     <div className='drag-line' style={{ borderTop: `${lineStyle} ${color} 1px` }} />
                 )}
                 <div className='draggable-area' />
-                <div className='drag-price' style={{ backgroundColor: color, width, opacity }}>
+                <div className='drag-price' style={{ backgroundColor: color, width, opacity, display: isBarrierZero ? 'none' : 'block' }}>
```
balakrishna-deriv, line 79: Can we prevent rendering this element instead of using CSS to hide it?
```
{ !isBarrierZero && <div className=....
```

akmal-deriv: Let me try that

balakrishna-deriv (approved):

src/components/PriceLine.tsx
```diff
@@ -50,14 +51,21 @@ const PriceLine = ({
     );
     const opacity = React.useMemo(() => (isOverlapping ? opacityOnOverlap : ''), [isOverlapping, opacityOnOverlap]);
 
+    const isBarrierZero = price === '+0.00';
```
balakrishna-deriv, line 54:
```suggestion
    const isBarrierZero = parseInt(price) === 0;
```

akmal-deriv: This can't parse integer since I need the value to be 0.00, so I'll pass float instead

henry-deriv (approved):

DerivFE (approved):

balakrishna-deriv (dismissed):

matin-deriv (approved):

matin-deriv (dismissed):

src/components/PriceLine.tsx
```diff
@@ -50,14 +51,21 @@ const PriceLine = ({
     );
     const opacity = React.useMemo(() => (isOverlapping ? opacityOnOverlap : ''), [isOverlapping, opacityOnOverlap]);
 
+    const isBarrierZero = parseFloat(price) === 0.00;
```
maryia-deriv, line 54:
```suggestion
    const isBarrierZero = parseFloat(price) === 0;
```
@akmal-deriv  even if you define a number with trailing zeros, they're dropped :) 
Screenshot 2023-09-08 at 8 54 34 AM

maryia-deriv (changes requested):

src/store/BarrierStore.ts
```diff
@@ -64,7 +65,7 @@ export default class BarrierStore {
         return this.mainStore.chart.yAxiswidth;
     }
     get priceLabelWidth(): number {
-        return this.yAxisWidth + 1;
+        return 16;
```
balakrishna-deriv, line 68: why are we hardcoding the width? It may not work properly for forex yaxis values as it might have more than 5 decimal places.

src/components/PriceLine.tsx
```diff
@@ -57,26 +58,51 @@ const PriceLine = ({
...
-                        <PriceLineArrow offScreenDirection={offScreenDirection} color={color} />
+                <div className='draggable-area-wrapper'>
+                    <div
+                        className={classNames('drag-price')}
+                        style={{ width: isOverlappingWithPriceLine ? width : yAxiswidth, opacity, right: isOverlappingWithPriceLine ? yAxiswidth - width : 0 }}
+                    >
+                        <div className='drag-icon'>
+                            <div></div>
+                            <div></div>
+                            <div></div>
```
balakrishna-deriv, line 94: 👀

balakrishna-deriv: why?

src/store/BarrierStore.ts
akmal-deriv: This number is only relevant for the case when barrier overlaps with current price. The value with all the decimals is moved out of this container and not affected by this width.

src/components/PriceLine.tsx
akmal-deriv: This is to generate a hamburger menu icon without the use of a symbol as I want the icon to be separated from the value inside.

src/store/BarrierStore.ts
balakrishna-deriv: Can we give a meaningful name? 😅

akmal-deriv: Sure, I replaced the name to something more relevant now.

balakrishna-deriv (approved):

src/store/BarrierStore.ts
```diff
@@ -83,12 +84,12 @@ export default class BarrierStore {
             hideOffscreenLine: observable,
             hideOffscreenBarrier: observable,
             isSingleBarrier: observable,
-            pip: computed,
-            yAxisWidth: computed,
-            priceLabelWidth: computed,
+            destructor: action.bound
```
kate-deriv, line 87: ` destructor: action.bound,`
 Comma :) Or it won't build

src/components/PriceLine.tsx
```diff
@@ -57,26 +58,51 @@ const PriceLine = ({
...
+                        </div>
+                        <div className={classNames('price', { 'price--zero': isOverlappingWithPriceLine })} style={{ right: isOverlappingWithPriceLine ? width + (priceDisplay.length * 8) - (!draggable ? 16 : 0) : 0}}>{priceDisplay}</div>
+                        <div />
```
kate-deriv, line 97: `!draggable ? 16 : 0 `===> `draggable ? 0 : 16`
But can be omitted)

kate-deriv: Or maybe it will make sense to lift up this calculation, like:
`const testVariable =  isOverlappingWithPriceLine ? width + (priceDisplay.length * 8) - (draggable ? 0 : 16) : 0}`
...
`style={{right: testVariable}}`

sass/components/_barrier.scss
```diff
@@ -123,26 +125,54 @@
...
-    padding: 3px 1px;
     line-height: 18px;
-    overflow: hidden;
+    overflow: unset;
+    padding: 3px 1px;
+    position: relative;
+    right: 45px;
     text-align: center;
-}
+    width: max-content;
```
kate-deriv, line 137: Cool) I never used this style before:)

sass/components/_barrier.scss
```diff
@@ -123,26 +125,54 @@
...
+        opacity: 0.3;
+        position: relative;
+        right: -10px;
+        top: -12px;
+    }
+}
 
-    &:after {
-        font-weight: normal;
-        content: '≡';
```
kate-deriv, line 140: Ahh, so you replaced this line with <div> x3? Nice)

src/components/PriceLine.tsx
akmal-deriv: I think it's fine, there's almost zero chance it will be reused anywhere else, so I didn't want to separate it. As for the first comment, I think having 0 as the first option is kind of counter-intuitive so I refer to put 0/null/undefined/etc options at the back)

src/store/BarrierStore.ts
akmal-deriv: Thanks, yes I sorted the lines, but it didn't have comma on one of it)

sass/components/_barrier.scss
akmal-deriv: This is one of the most useful things in CSS for me)

sass/components/_barrier.scss
akmal-deriv: Thanks, yes, the option with symbol is kind of old way and it was causing issues with moving the value out as it was a part of it

kate-deriv (approved):

balakrishna-deriv (dismissed):

src/components/PriceLine.tsx
markwylde-deriv: Would it make sense to abstract it into a `HamburgerDragIcon` component so it's clear?

henry-deriv (approved):

balakrishna-deriv (approved):

kate-deriv (approved):

src/components/PriceLine.tsx
```diff
@@ -57,26 +58,70 @@ const PriceLine = ({
...
+                    />
                 )}
                 <div className='draggable-area' />
-                <div className='drag-price' style={{ backgroundColor: color, width, opacity }}>
-                    <div className='price'>{priceDisplay}</div>
-                    {offScreen && offScreenDirection && (
-                        <PriceLineArrow offScreenDirection={offScreenDirection} color={color} />
+                <div className='draggable-area-wrapper'>
+                    <div
+                        className={classNames('drag-price')}
```
maryia-deriv, line 89:
```suggestion
                        className='drag-price'
```
@akmal-deriv nit-pick: just one classname :)

src/components/PriceLine.tsx
maryia-deriv: i agree with @markwylde-deriv

maryia-deriv (changes requested):
a bit late for the party, left a couple of comments, please check. the rest LGTM :)

src/components/PriceLine.tsx
akmal-deriv: Sure, fixed

src/components/PriceLine.tsx
akmal-deriv: Well, it's just one place where we used it, so I don't see any difference except for having more lines of code, so I didn't add it. And we'll probably never need to reuse this component anywhere else, since it's charts as well. But I moved it into a separate component since everybody doesn't like to see this in the file 😄

maryia-deriv (approved):

kate-deriv (approved):

src/components/PriceLine.tsx
markwylde-deriv: I think the difference is for documentation. The fact Bala had to ask what it's doing is, in my opinion, proof that it needs to be put into it's own component.

It doesn't necessarily need to go in it's own file.

PR #1419, opened 2023-06-14
Amir/feq 294/upgrade espree package version

amir-deriv:
## Summary

- https://github.com/eslint/espree/pull/546 confirms the support of node 18 in espree v9
- Ticker is working fine
- Menu switching working fine
- Changing language working fine
- No breaking change on v7 except the drop of node v8 support
- Breaking changes in the v8 with drop of node v10/13/15 support and other breaking changes is validated in the code
- Breaking change in v9 is the drop of using char  as a variable name since this is a resrved keyword in es3 but I found no usage of this reserved keyword in the code
- Removed tolerant  property from espree options since this is ignored in the package

### DIFF
--- a/scripts/extract-translations.js
+++ b/scripts/extract-translations.js
@@ -13,7 +13,6 @@ const default_options = {
 const parseCode = source => espree.parse(source, {
     ecmaVersion: 10,
     sourceType: 'script',
-    tolerant: true,
     loc: true,
 }).body;
 
--- a/scripts/extract-translations.js
+++ b/scripts/extract-translations.js
@@ -11,7 +11,7 @@ const default_options = {
 };
 
 const parseCode = source => espree.parse(source, {
-    ecmaVersion: 'latest',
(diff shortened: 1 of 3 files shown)


scripts/extract-translations.js
```diff
@@ -11,9 +11,8 @@ const default_options = {
 };
 
 const parseCode = source => espree.parse(source, {
-    ecmaVersion: 10,
+    ecmaVersion: 'latest',
```
ali-hosseini-deriv, line 14: hey @amir-deriv can you make this something static instead of latest?
asking this because in case of any update on ecma then it will use that, not that much safe

amir-deriv: Done @ali-hosseini-deriv 

image

The error was happening due to optional chaining '?' which is part of `es2020` so the `ecmaVersion` of it is `11` , I have updated it to `11` the latest version is `13` right now

PR #1434, opened 2023-08-17
Feature adding hover on indicator

ahmadtaimoor-deriv:


src/store/ChartAdapterStore.ts
```diff
@@ -58,15 +163,9 @@ export default class ChartAdapterStore {
...
-                // dxLocal and dyLocal are the local position value correponding to the bottom indicator/main chart
-                const epoch = this.flutterChart?.crosshair.getEpochFromX(dxLocal) || 0;
-                const quote = (this.flutterChart?.crosshair.getQuoteFromY(dyLocal) || 0).toFixed(
-                    this.mainStore.crosshair.decimalPlaces
-                );
-
-                this.mainStore.crosshair.onMouseMove(dx, dy, epoch, quote);
+            onCrosshairHover: (dx, dy, dxLocal, dyLocal, bottomIndicatorIndex) => {
+                const debounced = this.debounce(this.crossHover, 100);
+                debounced(dx, dy, dxLocal, dyLocal, bottomIndicatorIndex);
```
balakrishna-deriv, line 168: can you use debounce from 'lodash.debounce'? We already have that in our package.json

balakrishna-deriv: Let's not do debounce for tooltip move. Tooltip won't take that much computing and it has to be smooth.
Can we do debounce only for indicator hover check?

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@deriv/deriv-charts",
-    "version": "1.3.2",
-    "lockfileVersion": 3,
+    "version": "1.3.1",
+    "lockfileVersion": 2,
```
balakrishna-deriv, line 4: why are we changing this here?

ahmadtaimoor-deriv: fixed now , it got changed during merge conflict fixes , Thanks

src/store/ChartAdapterStore.ts
ahmadtaimoor-deriv: lodesh.debounce is used now . Debounce for hover is only done now , thanks

balakrishna-deriv (approved):

PR #1439, opened 2023-08-28
Maryia/dtra-369/Send chart events using stateChangeListener for analytics collection

maryia-deriv:


src/store/ChartState.ts
```diff
@@ -739,6 +754,10 @@ class ChartState {
             this.stxx.chart.lockScroll = false;
         }
     }
+
+    debouncedStateChange = debounce((state: string, option: TStateChangeOption) => {
+        this.stateChange(state, option);
+    }, 500);
```
maryia-deriv, line 760: waiting for user to stop typing

src/components/ChartTitle.tsx
```diff
@@ -68,9 +68,14 @@ const ChartTitle = (props: TChartTitleProps) => {
...
                     searchInputClassName={searchInputClassName}
-                    onSelectItem={(x: TProcessedSymbolItem) => {
-                        if (x.symbol !== currentSymbol.symbol) {
-                            onChange(x.symbol);
+                    onSelectItem={(symbol_object: TProcessedSymbolItem, category_id: string) => {
+                        if (symbol_object.symbol !== currentSymbol.symbol) {
+                            onChange(symbol_object.symbol);
+                            state.stateChange(STATE.SYMBOL_CHANGE, {
+                                tab_market_name:
+                                    category_id === 'favorite' ? 'favorites' : getSymbolMarketCategory(symbol_object),
```
henry-deriv, line 76: Referring to this it could also be `symbol_market_category` name instead of `tab_market_name` https://github.com/binary-com/deriv-app/pull/9848/files#r1316656672

src/components/ChartTitle.tsx
```diff
@@ -68,9 +68,14 @@ const ChartTitle = (props: TChartTitleProps) => {
...
+                                market_type_name: symbol_object.symbol,
```
henry-deriv, line 77: im just wondering why are we setting market_type_name to symbol name?

src/store/CategoricalDisplayStore.ts
```diff
@@ -343,6 +344,7 @@ export default class CategoricalDisplayStore {
         if (filterText === '') {
             setTimeout(() => this.scrollToActiveSymbol(), 1);
         }
+        this.mainStore.state.debouncedStateChange(STATE.MARKET_SEARCH, { search_string: filterText });
```
henry-deriv, line 347: the purpose of this is to send statechange when user stops typing for a while right? instead of sending an state/event at every character change

src/store/ChartState.ts
henry-deriv: nice. Just saw after i commented

src/store/CategoricalDisplayStore.ts
maryia-deriv: @henry-deriv yes, correct :)

src/components/ChartTitle.tsx
maryia-deriv: @henry-deriv it's because we need to send it in this form to rudderstack via deriv-app. I can either call this property straight the way it needs to go to analytics and have less logic in deriv-app - it's what I do know, or name it in the way we want here and rename this properties later when they are received inside deriv-app.

henry-deriv: sorry. i was under the impression that market_type_name referrers to derived, forex, cryptocurrencies, etc. Thats why i was confused as to why symbol is passed as market_type_name

Screenshot 2023-09-06 at 4 14 23 PM

kate-deriv (approved):

src/components/ChartTitle.tsx
maryia-deriv: @henry-deriv marketing team has their own names for all this :)) hm, I think I can change it here, and rename the property in deriv-app PR before we send it to rudderstack

henry-deriv: sounds good

kate-deriv (approved):

akmal-deriv (approved):

akmal-deriv (approved):

PR #1446, opened 2023-09-08
Kate / DTRA-383 / Improvement/ On the Chart "price info box", date is displayed in mm/dd format

kate-deriv:
- Change date format in chart info-box (mm/dd ---> to dd/mm)
- Change check for rendering symbol field in chart info-box

### DIFF
--- a/src/store/CrosshairStore.ts
+++ b/src/store/CrosshairStore.ts
@@ -380,7 +380,7 @@ class CrosshairStore {
                         if (stx.chart.xAxis.noDraw) {
                             continue;
                         } else {
-                            const formattedTime = floatDate.innerHTML.replace(/(\d{2})\/(\d{2})/, '$2/$1,');
+                            const formattedTime = floatDate.innerHTML.replace(/(\d{2})\/(\d{2})/, '$2/$1');
                             fieldValue = formattedTime;
                         }
                     } else {
--- a/src/store/CrosshairStore.ts
+++ b/src/store/CrosshairStore.ts
@@ -1,6 +1,5 @@
 import { action, computed, observable, when, makeObservable } from 'mobx';
 import Context from 'src/components/ui/Context';
-import moment from 'moment';
 import { TQuote } from 'src/types';
 import MainStore from '.';
 import Theme from '../../sass/_themes.scss';
@@ -381,8 +380,8 @@ class CrosshairStore {
                         if (stx.chart.xAxis.noDraw) {
                             continue;
                         } else {
-                            const formattedTime = floatDate.innerHTML.split(' ');
-                            fieldValue = `${moment(formattedTime[0], 'MM/DD').format('DD/MM')} ${formattedTime[1]}`;
+                            const formattedTime = floatDate.innerHTML.replace(/(\d{2})\/(\d{2})/, '$2/$1');
+                            fieldValue = formattedTime;
                         }
                     } else {
                         fieldValue = CIQ.yyyymmdd(dsField);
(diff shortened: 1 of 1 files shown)


src/store/CrosshairStore.ts
```diff
@@ -236,7 +236,7 @@ class CrosshairStore {
                 dupMap.Open = dupMap.High = dupMap.Low = 1;
             }
         }
-        if (this.showSeries) {
+        if (this.showSeries && this.mainStore.timeperiod.timeUnit === 'tick') {
```
kate-deriv, line 239: `this.showSeries` is equal to `true`, so I tried to add `this.mainStore.timeperiod.timeUnit === 'tick'` instead of default value, but failed as `mainStore` is not initialised there. I'm thinking of removing `this.showSeries` at all from this file as it's used only in this line. Please, share your thoughts, as it's scary to delete smth from SmartCgart)
Screenshot 2023-09-11 at 1 11 02 PM

src/store/CrosshairStore.ts
```diff
@@ -380,7 +380,8 @@ class CrosshairStore {
                         if (stx.chart.xAxis.noDraw) {
                             continue;
                         } else {
-                            fieldValue = floatDate.innerHTML;
+                            const formattedTime = floatDate.innerHTML.replace(/(\d{2})\/(\d{2})/, '$2/$1');
```
kate-deriv, line 383: I have no access to file with function, where the format time function is located (`chartiq`), so I added modification here (replace mm/dd ---> to dd/mm)

src/store/CrosshairStore.ts
maryia-deriv: @kate-deriv please elaborate on how this change is related to the date format in the price info box? :) If we need to only change the format of the date for all possible cases, we don't need to change this line, do we?

src/store/CrosshairStore.ts
maryia-deriv: @kate-deriv  let's try using 'moment' library to format the date in a way we want :)

maryia-deriv (changes requested):

src/store/CrosshairStore.ts
kate-deriv: @maryia-deriv It is not related to the date format, it is related to removing symbol field in chart info-box for non-ticks time intervals (please, check the notes in the card, this PR is common for 2 cards)

src/store/CrosshairStore.ts
kate-deriv: Will do, thank you

kate-deriv: I implemented solution with moment library, but found out, that format of date depends on time intervals. For ticks: it's _mm/dd hh:mm:ss_, for minutes and hours: _mm/dd hh:mm_, for days:` mm/dd/yyyy`. So in order to use moment library I need add different conditions and checks. 
What doesn't change in date is the first 5 symbols in the string: _mm/dd_. That is why usage of reg exp (which just swap first numbers) is more minimalistic solution.

maryia-deriv (approved):

henry-deriv (approved):

src/store/CrosshairStore.ts
```diff
@@ -236,7 +236,7 @@ class CrosshairStore {
                 dupMap.Open = dupMap.High = dupMap.Low = 1;
             }
         }
-        if (this.showSeries) {
+        if (this.showSeries && !this.showOhl) {
```
balakrishna-deriv, line 239: Why have we added `!this.showOhl` check? 
OHLC chart type should also have the same date format

kate-deriv: Hi @balakrishna-deriv ! This is not related to date format, it is related to removing symbol field in chart info-box for non-ticks time intervals (please, check the notes in the card, this PR is common for 2 cards)

balakrishna-deriv (approved):

matin-deriv (approved):

PR #1453, opened 2023-10-11
[DTRA] Maryia/DTRA-411/fix: chart lagging after applying a saved layout template

maryia-deriv:
This includes the following fixes:
1) Fixing the chart lagging issue, which was happening when a user applies a saved template, by ensuring that we load a template after the chart is ready, and set the correct granularity.
Now, when another chart mode is set upon application of a template:
    - We notify the parent about the chart mode change using `stateChange` callback,
    - We set granularity & chart type.

2) Now, when only a 1-tick "Area" chart mode is allowed:
    - `InfoFootnote` is displayed in the Templates modal informing a user that not all templates can be applied.
    - Templates will be sorted to display: enabled (applicable) ones first, disabled (inapplicable) ones in the end.
   
        <image src='https://github.com/binary-com/SmartCharts/assets/103177211/621295b1-795c-4382-83ba-30d9b0f17cba' width='500px'/>

3) Removing inapplicable indicators from the chart when users switch to 1-Tick Area chart for which only some indicators are available.


src/store/ViewStore.ts
```diff
@@ -172,12 +175,14 @@ export default class ViewStore {
...
                 this.mainStore.state.setChartIsReady(true);
                 this.mainStore.state.setChartGranularity(granularity);
+                importFinishedPromise.resolve();
             };
             stx.importLayout(sortedItems, {
                 managePeriodicity: true,
                 preserveTicksAndCandleWidth: true,
                 cb: finishImportLayout,
             });
+            await importFinishedPromise;
```
maryia-deriv, line 185: Using this promise to ensure that the code below is executed not before but after `finishImportLayout` is executed. The issue happened because the order of execution was random.

src/store/ViewStore.ts
```diff
@@ -159,10 +160,12 @@ export default class ViewStore {
         this.mainStore.state.setChartIsReady(false);
         const sortedItems = this.sortedItems[idx].layout;
         const stx = this.stx;
-        const granularity = getIntervalInSeconds(sortedItems) as TGranularity;
+        const { interval, periodicity, timeUnit } = sortedItems;
+        const period = timeUnit ? interval : periodicity;
+        const granularity = calculateGranularity(Number(period), (timeUnit as string) || (interval as string));
 
-        this.mainStore.timeperiod.onGranularityChange(granularity);
```
maryia-deriv, line 164: we don't need to call this callback since it's not assigned with any function unless we open Chart mode modal to change Chart type of Time interval before applying a template.
`this.mainStore.state.setChartGranularity(granularity);` at L177 serves this purpose well enough.

src/store/ViewStore.ts
```diff
@@ -159,10 +160,12 @@ export default class ViewStore {
         this.mainStore.state.setChartIsReady(false);
         const sortedItems = this.sortedItems[idx].layout;
         const stx = this.stx;
-        const granularity = getIntervalInSeconds(sortedItems) as TGranularity;
+        const { interval, periodicity, timeUnit } = sortedItems;
+        const period = timeUnit ? interval : periodicity;
+        const granularity = calculateGranularity(Number(period), (timeUnit as string) || (interval as string));
```
maryia-deriv, line 165: One of the reasons why the issue was happening was `getIntervalInSeconds()` returning a wrong `granularity` value.

src/components/Views.tsx
```diff
@@ -91,10 +108,11 @@ const ActiveListView = ({ views, removeAll, applyLayout, remove }: TActiveListVi
             </div>
             <div className='sc-views__views__content'>
                 <div className='sc-views__views__list'>
-                    {views.map((view, i) => (
+                    {sorted_views.map((view, i) => (
                         <ViewItem
+                            disabled={isDisabled(view.layout)}
                             view={view}
-                            key={view.name}
+                            key={`${view.name}-${i.toString()}`}
```
maryia-deriv, line 115: `view.name` is set by a user, so it can be the same for more than 1 template. we need a unique key here, hence adding a stringified index for this purpose.

src/store/ChartState.ts
```diff
@@ -510,6 +514,9 @@ class ChartState {
         if (this.context && isTimeUnitSecond && isChartTypeCandle) {
             this.setChartType('mountain'); // if granularity is zero, set the chartType to 'mountain'
         }
+        if (granularity === 0 && this.mainStore.studies.hasPredictionIndicator) {
+            this.mainStore.studies.deletePredictionStudies();
```
maryia-deriv, line 518: Helps clear those indicators that are inapplicable for a chart mode that we are switching to.

sass/components/_dialog.scss
```diff
@@ -132,6 +132,20 @@
             }
         }
     }
+    .info-footnote {
+        @include themify($themes) {
+            color: themed('ChartModeText') !important;
+        }
+        line-height: 18px;
```
kate-deriv, line 139: Just asking: how we determine when to use `rem` and `px`?

src/components/InfoFootnote.tsx
```diff
@@ -0,0 +1,25 @@
...
+type InfoFootnoteProps = {
+    className?: string;
+    isMobile?: boolean;
+    text: string;
+};
+
+const InfoFootnote = ({ className, isMobile, text }: InfoFootnoteProps) => (
+    <div
+        className={classNames('info-footnote', className, {
+            [`${className}--mobile`]: isMobile,
```
kate-deriv, line 13: Just asking: can't we use `@include mobile` in `.css` file instead of passing `isMobile` and adding `className`?

src/components/Views.tsx
```diff
@@ -28,24 +30,29 @@ type TOverwriteViewProps = {
...
     applyLayout: TMainStore['view']['applyLayout'];
     remove: TMainStore['view']['remove'];
 };
 
-const ViewItem = ({ view, remove, onClick }: TViewItemProps) => (
+const ViewItem = ({ disabled, view, remove, onClick }: TViewItemProps) => (
     <Tooltip
         className='sc-views__views__list__item'
-        onClick={onClick}
+        onClick={disabled ? undefined : onClick}
```
kate-deriv, line 43: Nice trick to prevent from clicking on disable)

src/components/Views.tsx
```diff
@@ -78,8 +85,18 @@ const OverwriteView = ({ templateName, onCancel, onOverwrite }: TOverwriteViewPr
...
 
-const ActiveListView = ({ views, removeAll, applyLayout, remove }: TActiveListViewProps) => {
+const ActiveListView = ({ allowTickChartTypeOnly, views, removeAll, applyLayout, remove }: TActiveListViewProps) => {
     if (!views.length) return null;
+    const isDisabled = (layout: TMainStore['view']['sortedItems'][number]['layout']) => {
+        const { chartType, periodicity, timeUnit } = layout ?? {};
+        const oneTickChartTemplate = chartType === 'mountain' && periodicity === 1 && timeUnit === 'second';
+        return allowTickChartTypeOnly && !oneTickChartTemplate;
+    };
+    const sorted_views = [...views].sort((a, b) => {
```
kate-deriv, line 95: if you don't want to mutate `views` array, you can use `.toSorted()` method (`.toSorted()` method of Array instances is the copying version of the `.sort()` method; it returns a new array with the elements sorted) instead of creating copy and sorting it

src/components/Views.tsx
kate-deriv: I didn't get the idea of stringified index. If it's number and it's inserted in backticks, won't it be a part of the string? Just curious

src/components/Views.tsx
kate-deriv: A got it, it's not supported in SmartCharts

kate-deriv (approved):

src/components/ChartMode.tsx
```diff
@@ -66,17 +67,18 @@ const ChartMode = ({ onChartType, onGranularity, portalNodeId }: TChartModeProps
...
+                        isMobile={isMobile}
+                        text={t.translate('Only selected charts and time intervals are available for this trade type.')}
+                    />
                 )}
             </Menu.Body>
         </Menu>
     );
 };
 
+ChartMode.defaultProps = {
```
henry-deriv, line 80: cant we just initialize instead of doing this

src/components/InfoFootnote.tsx
```diff
@@ -0,0 +1,25 @@
...
+        })}
+    >
+        {text}
+    </div>
+);
+
+InfoFootnote.defaultProps = {
```
henry-deriv, line 20: same here as above

src/utils/__tests__/index.spec.ts
```diff
@@ -149,3 +155,163 @@ describe('getTimeIntervalName', () => {
...
+    it('should return correct granularity for 1 tick', () => {
+        expect(calculateGranularity(1, 'second')).to.equal(0);
+    });
+    it('should return correct granularity for 1 minute', () => {
+        expect(calculateGranularity(1, 'minute')).to.equal(60);
+    });
+    it('should return correct granularity for 2 minutes', () => {
+        expect(calculateGranularity(2, 'minute')).to.equal(120);
+    });
+    it('should return correct granularity for 3 minutes', () => {
```
akmal-deriv, line 169: Do we really need so many almost identical test cases? The chance that it would succeed for say 1,2,3 minutes and fail for 5 minutes is very low but each test takes time to execute what makes our deployment take longer time.

akmal-deriv (changes requested):

src/utils/__tests__/index.spec.ts
maryia-deriv: @akmal-deriv please clarify what you mean. :) It's just smartcharts repo, not deriv-app, and it's launched very quickly with npm start. And when you run tests there are only 68 test cases in the app in total that are passing within 34ms. Also, having these test cases for each case can help us see which input exactly will result in which output when we use these functions.

src/components/ChartMode.tsx
maryia-deriv: Screenshot 2023-10-23 at 6 28 33 PM
i tried it but it didn't help. it's about project settings.

src/components/InfoFootnote.tsx
maryia-deriv: Screenshot 2023-10-23 at 6 29 38 PM
same :)

henry-deriv (approved):

src/utils/__tests__/index.spec.ts
akmal-deriv: I don't think adding similar test cases would anyhow improve developers experience, we're adding test cases to make sure the function/component work as expected and adding similar cases doesn't make any difference as they are almost identical. They would either all pass or all fail at the same time. 😄 But it's ok anyways.

akmal-deriv (approved):

src/utils/__tests__/index.spec.ts
maryia-deriv: thank you @akmal-deriv )
yeah, i just want to use these test cases for presentational purpose in order to reduce time for debugging in future ;)

PR #1462, opened 2023-11-28
Akmal / feat: implement new blue barrier in flutter charts

akmal-deriv:
This pull request (PR) introduces a new feature, specifically the implementation of a new blue barrier, in the Flutter charts. The term "barrier" in this context likely refers to a graphical element within the chart, and the addition of a "blue barrier" suggests the introduction of a new visual component with a blue color. This feature implementation is intended to enhance or extend the functionality of the Flutter charts, providing users with the ability to incorporate and interact with the newly introduced blue barrier.

### DIFF
--- a/chart_app/lib/src/helpers/chart.dart
+++ b/chart_app/lib/src/helpers/chart.dart
@@ -30,7 +30,7 @@ double calculateCurrentTickWidth(
       pipSize,
     );
 
-    return width + 4 * 3;
+    return width + 4 * 2 + 3;
   }
 }
 
--- a/chart_app/lib/src/helpers/chart.dart
+++ b/chart_app/lib/src/helpers/chart.dart
@@ -1,6 +1,7 @@
 import 'dart:math';
 
 import 'package:deriv_chart/deriv_chart.dart';
+import 'package:flutter/material.dart';
 
 /// Calculates the width of yAxis
 double calculateYAxisWidth(List<Tick> ticks, ChartTheme theme, int pipSize) {
@@ -8,12 +9,28 @@ double calculateYAxisWidth(List<Tick> ticks, ChartTheme theme, int pipSize) {
     return 60;
   } else {
     final double width = labelWidth(
-      ticks.first.close,
+      ticks.last.close,
       theme.gridStyle.yLabelStyle,
       pipSize,
     );
 
     return width + theme.gridStyle.labelHorizontalPadding * 2;
+  }
+}
+
+/// Calculates the width of current tick
+double calculateCurrentTickWidth(
+    List<Tick> ticks, TextStyle textStyle, int pipSize) {
+  if (ticks.isEmpty) {
+    return 60;
+  } else {
+    final double width = labelWidth(
+      ticks.last.close,
+      textStyle,
+      pipSize,
+    );
+
+    return width + 4 * 3;
   }
 }
 
--- a/sass/components/_barrier.scss
+++ b/sass/components/_barrier.scss
@@ -1,12 +1,13 @@
 @charset "UTF-8";
 
 .barrier-area {
+    -moz-user-select: none;
+    -webkit-user-select: none;
     padding: 13px 0;
     position: absolute;
-    width: 100%;
+    top: 0;
     user-select: none;
-    -webkit-user-select: none;
-    -moz-user-select: none;
+    width: 100%;
     z-index: 1;
 }
 /* Drag Lines and Prices */
--- a/sass/components/_barrier.scss
+++ b/sass/components/_barrier.scss
@@ -40,7 +40,8 @@
             width: 100%;
             margin-left: -10px;
             position: relative;
-            border-top-width: 0;
+            border-top-width: 1px;
+            border-top-color: $color-blue;
         }
         .title-wrapper {
             $title-height: 24px;
@@ -87,13 +88,13 @@
 }
 
 .drag-price {
-    display: block;
-    height: 23px;
-    top: -11px;
+    display: flex;
+    height: 24px;
+    top: -12px;
     position: absolute;
     right: -1px;
-    border: 1px solid inherit;
-    user-select: none;
+    background-color: $color-blue;
+    justify-content: space-between;
 
     .arrow-icon {
         height: 41px;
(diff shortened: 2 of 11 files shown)


src/components/PriceLine.tsx
```diff
@@ -54,32 +58,72 @@ const PriceLine = ({
...
+                        >
+                            {priceDisplay}
+                        </div>
+                        <div />
+                        {offScreen && offScreenDirection && (
+                            <PriceLineArrow offScreenDirection={offScreenDirection} color={color} />
+                        )}
+                    </div>
+                    {isOverlappingWithPriceLine && (
+                        <div>
```
kate-deriv, line 118: Sorry if it's irrelevant, do we need extra empty `<div>` tag here?

src/components/PriceLine.tsx
```diff
@@ -54,32 +58,72 @@ const PriceLine = ({
...
     if (!showBarrier) return null;
 
-    const width = showBarrierDragLine ? priceLineWidth + 16 : priceLineWidth;
+    const width = priceLineWidth + 12;
 
     return (
-        <div className='barrier-area' style={{ top: 0 }} ref={setDragLine} hidden={!visible}>
+        <div
+            className={classNames('barrier-area', { 'barrier-area--zero': isOverlappingWithPriceLine })}
+            style={{ top: 0 }}
```
kate-deriv, line 66: if it has `className`, will it make sense to move `style={{ top: 0 }} `inside of css?

src/store/PriceLineStore.ts
```diff
@@ -175,8 +178,11 @@ export default class PriceLineStore {
     }
 
     get priceLineWidth() {
-        // TODO: measure the price label width
-        return Math.max(this.priceDisplay.length * 8, 60);
+        return window.flutterChart?.app.getCurrentTickWidth() || 60;
+    }
+
+    get overlappedBarrierWidth(): number {
+        return 16;
```
kate-deriv, line 185: if this getter always returns 16, maybe it'll make sense to create a constant?

chart_app/lib/src/helpers/chart.dart
```diff
@@ -17,6 +18,22 @@ double calculateYAxisWidth(List<Tick> ticks, ChartTheme theme, int pipSize) {
   }
 }
 
+/// Calculates the width of current tick
```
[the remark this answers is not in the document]
balakrishna-deriv: It's mandatory to add a comment in Dart if is a public function/variable. There is a flutter lint to check this.

src/store/PriceLineStore.ts
akmal-deriv: Yeah, but I don't want to change this as there might be a need to make this dynamic again in the future and someone will have to figure it out and write it up again from scratch, so let's leave it here.

src/store/BarrierStore.ts
```diff
@@ -61,26 +62,31 @@ export default class BarrierStore {
         return this.mainStore.chart.currentActiveSymbol?.decimal_places as number;
     }
 
+    get overlappedBarrierWidth(): number {
```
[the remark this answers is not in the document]
akmal-deriv: Indeed, seems like it's a duplicate in a wrong store 👍🏻

chart_app/lib/src/helpers/chart.dart
```diff
@@ -17,6 +18,22 @@ double calculateYAxisWidth(List<Tick> ticks, ChartTheme theme, int pipSize) {
...
+  if (ticks.isEmpty) {
+    return 60;
+  } else {
+    final double width = labelWidth(
+      ticks.last.close,
+      textStyle,
+      pipSize,
+    );
+
+    return width + 4 * 2 + 3;
```
henry-deriv, line 33: why not just return 11?

akmal-deriv: This width is based on calculation, it makes more sense to break it down than return meaningless value that would be hard to change as well as it wouldn't be clear what it's based on.

src/components/PriceLine.tsx
akmal-deriv: I removed it, looks like it's not in use anymore

chart_app/lib/src/helpers/chart.dart
henry-deriv: in that case better to name them as constants so we know what each number stands for

akmal-deriv: Well, it doesn't really stand for anything, therefore I didn't name it, it's just approximate symbol width (4) and the rest are just random multiplier and padding required to make it work. What I mean is it's same as just use + 11 but something which makes better visual representation and at least we can see it's based on some kind of calculation. I don't think I can come up with any names for these numbers.

kate-deriv (approved):

balakrishna-deriv (approved):

henry-deriv (approved):

maryia-deriv (approved):

src/components/PriceLine.tsx
```diff
@@ -54,32 +58,69 @@ const PriceLine = ({
...
+                            backgroundColor: color,
+                            width: isOverlappingWithPriceLine ? overlappedBarrierWidth : width,
+                            opacity,
+                            right: (isOverlappingWithPriceLine ? width - overlappedBarrierWidth : 0) + (isMobile ? 20 : 4),
+                        }}
+                    >
+                        <HamburgerDragIcon />
+                        <div
+                            className={classNames('price', { 'price--zero': isOverlappingWithPriceLine })}
+                            style={{
```
ahmadtaimoor-deriv, line 102: Can we reduce the ternary operator by something like 

style={isOverlappingWithPriceLine?{} :{} }

sass/components/_barrier.scss
```diff
@@ -87,13 +89,13 @@
 }
 
 .drag-price {
-    display: block;
-    height: 23px;
-    top: -11px;
+    display: flex;
+    height: 24px;
```
ahmadtaimoor-deriv, line 93: Just asking , we don't follow rem instead of px in smartcharts?

akmal-deriv: No, in deriv-app rem is fixed to 10px, while in smartcharts it's not, so it would be given a dynamic value which various on different devices, which we don't want;

ahmadtaimoor-deriv (approved):

balakrishna-deriv (approved):

PR #1476, opened 2024-01-15
Bala/Fix trackjs issues

balakrishna-deriv:
**Changelog**

To fix a trackjs issue caused by some keyboard events that don't contain `key` or `code` props.
https://github.com/flutter/engine/blob/f20657354d8b53baafcec55650830ead89adf3e9/lib/web_ui/lib/src/engine/keyboard_binding.dart#L386

### DIFF
--- a/src/flutter-chart/index.ts
+++ b/src/flutter-chart/index.ts
@@ -12,6 +12,8 @@ export const createChartElement = ({ onChartLoad }: TCreateChartElementProps) =>
         return;
     }
 
+    setupListeners();
+
     const flutterChartElement = document.createElement('div');
     flutterChartElement.classList.add('flutter-chart');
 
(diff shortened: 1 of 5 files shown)


src/flutter-chart/index.ts
```diff
@@ -42,6 +44,19 @@ export const createChartElement = ({ onChartLoad }: TCreateChartElementProps) =>
...
+const setupListeners = () => {
+    const listener = (ev: KeyboardEvent) => {
+        // To fix a trackjs issue caused by some keyboard events that don't contain `key` or `code` props.
+        // https://github.com/flutter/engine/blob/f20657354d8b53baafcec55650830ead89adf3e9/lib/web_ui/lib/src/engine/keyboard_binding.dart#L386
+        if (!ev.key || !ev.code) {
+            ev.stopImmediatePropagation();
+        }
+    };
+
+    window.addEventListener('keydown', listener, true);
```
kate-deriv, line 56: Do we need to clean (remove) this eventListeners somewhere?  They will be added on each call of the createChartElement()

balakrishna-deriv: Once flutter is loaded we always keep it in the background to make the chart load fast when it is loaded again. So, we shouldn't remove these two event listeners.

akmal-deriv (approved):

ahmadtaimoor-deriv (approved):

kate-deriv (approved):

henry-deriv (approved):

PR #1510, opened 2024-02-16
[DTRA]/Ahmad/FEQ-970/Template fix

ahmadtaimoor-deriv:
# Description
This PR includes the fixes required to make the templates feature work in the Chart. It ensures that templates are working now with respect to time interval, chart type, indicators and crosshair. It allows on both bot and trader charts.


app/test.tsx
```diff
@@ -1,902 +0,0 @@
...
-import { TNotification } from 'src/store/Notifier';
-import {
-    TBarrierChangeParam,
-    TGranularity,
-    TInitialChartData,
-    TNetworkConfig,
-    TOpenMarket,
-    TRefData,
-    TSettings,
-} from 'src/types';
```
ahmadtaimoor-deriv, line 31: Confirmed from bala that we don't use this file anymore and can get rid of it

app/multichart.tsx
```diff
@@ -1,100 +0,0 @@
...
-    Share,
-} from '@binary-com/smartcharts'; // eslint-disable-line import/no-unresolved
-import React from 'react';
-import ReactDOM from 'react-dom';
-import './test.scss';
-import { ConnectionManager, StreamManager } from './connection';
-
-setSmartChartsPublicPath('./dist/');
-
-configure({ enforceActions: 'observed' });
```
ahmadtaimoor-deriv, line 25: we don't use this file anymore and can get rid of it. Confirmed from bala

src/components/ChartControls.tsx
```diff
@@ -22,7 +21,6 @@ export const RenderDefaultControls = ({ isMobile }: TRenderDefaultControls) => (
         <Timeperiod />
         <StudyLegend />
         <DrawTools />
-        <Views />
```
ahmadtaimoor-deriv, line 25: ChartControls is called in Chart.tsx but 
we always pass chartControlsWidgets as null from deriv app. So it never get called.

```
 {chartControlsWidgets !== null && !enabledChartFooter && (
                                <ChartControls widgets={chartControlsWidgets} />
                            )}
```

farhan-nurzi-deriv (approved):

akmal-deriv (approved):

henry-deriv (approved):

src/components/Views.tsx
```diff
@@ -19,6 +19,8 @@ type TViewItemProps = {
 
 type TViewsProps = {
     portalNodeId?: string;
+    onChartType: (chartType?: string | undefined) => void;
```
kate-deriv, line 22: Is this extra undefined needed here?
`onChartType: (chartType?: string) => void;`

src/components/Views.tsx
```diff
@@ -32,6 +34,8 @@ type TActiveListViewProps = {
     views: TMainStore['view']['sortedItems'];
     applyLayout: TMainStore['view']['applyLayout'];
     remove: TMainStore['view']['remove'];
+    onChartType: (chartType?: string | undefined) => void;
```
kate-deriv, line 37: Same here

src/store/CrosshairStore.ts
```diff
@@ -97,17 +99,24 @@ class CrosshairStore {
...
             this.isOverChartContainer = true;
-            this.updateVisibility(true);
             this.hoverOnScreen = true;
         }
-
         if (!this.isOverChartContainer) return;
 
+        if (this.layoutCrosshair) {
+            this.setCrosshairState(this.layoutCrosshair);
+            this.layoutCrosshair = 0;
```
kate-deriv, line 115: Why do we need to call a function  `this.setCrosshairState`, if then on line 115 we set it to zero?
Literally it seems like:
```
this.layoutCrosshair = this.layoutCrosshair;
this.layoutCrosshair = 0;
```
Sorry I didn't get the idea :( Please, ignore if it's not relevant

src/store/ViewStore.ts
```diff
@@ -149,17 +149,22 @@ export default class ViewStore {
     removeAll() {
         this.views = [];
         this.updateLocalStorage();
-        logEvent(LogCategories.ChartControl, LogActions.Template, 'Remove All Templates');
+        logEvent(LogCategories.ChartControl, LogActions.Template,'Remove All Templates');
         this.updateRoute('new');
     }
 
-    applyLayout(idx: number, e: TCustomEvent) {
+    applyLayout(idx: number, e: TCustomEvent,onGranularity:(granularity?: TGranularity) => void,onChartType: (chartType:string|undefined) => void) {
```
kate-deriv, line 156: Will it work?
`onChartType: (chartType?:string) => void`

src/store/CrosshairStore.ts
ahmadtaimoor-deriv: So layoutCrosshair is crosshair value that is set up from the template . Once the template is clicked . It gets the crosshair of the template into layoutCrosshair and not on the setCrosshairState(the actual crosshair state) . It only set it up on   onmousemove  function.
The above code . It checks if there is a layoutCrosshair state . If yes it sets up the actual crosshair state (inside setCrosshairState function) from layoutCrosshair and then make its value 0 until a new template is set up.

kate-deriv: Thank you for explanation

kate-deriv (approved):

src/store/ViewStore.ts
```diff
@@ -149,17 +149,22 @@ export default class ViewStore {
     removeAll() {
         this.views = [];
         this.updateLocalStorage();
-        logEvent(LogCategories.ChartControl, LogActions.Template, 'Remove All Templates');
+        logEvent(LogCategories.ChartControl, LogActions.Template,'Remove All Templates');
         this.updateRoute('new');
     }
 
-    applyLayout(idx: number, e: TCustomEvent) {
+    applyLayout(idx: number, e: TCustomEvent,onGranularity:(granularity?: TGranularity) => void,onChartType: (chartType?:string) => void) {
```
maryia-deriv, line 156: @ahmadtaimoor-deriv it looks like space is missing here before `onGranularity`, before `onChartType`, & before `string`. Could you please run option+shift+f in this file to prettify it ?
Also, i noticed that you're removing `multichart` and `test.tsx` in this PR but you're doing the same in another JS cleanup PR. This will cause a conflict. Why do we need to remove multichart and test.tsx in this PR?

src/store/ViewStore.ts
```diff
@@ -149,17 +149,22 @@ export default class ViewStore {
     removeAll() {
         this.views = [];
         this.updateLocalStorage();
-        logEvent(LogCategories.ChartControl, LogActions.Template, 'Remove All Templates');
+        logEvent(LogCategories.ChartControl, LogActions.Template,'Remove All Templates');
```
maryia-deriv, line 152: why do we remove the space before 'Remove All... ?

src/components/Views.tsx
```diff
@@ -95,7 +99,7 @@ const ActiveListView = ({ views, removeAll, applyLayout, remove }: TActiveListVi
                         <ViewItem
                             view={view}
                             key={view.name}
-                            onClick={e => applyLayout(i, e as TCustomEvent)}
+                            onClick={e => applyLayout(i, e as TCustomEvent,onGranularity,onChartType)}
```
maryia-deriv, line 102: spaces are missing, please prettify files of this PR

maryia-deriv (changes requested):

maryia-deriv (approved):

kate-deriv (approved):

maryia-deriv (approved):

PR #1521, opened 2024-03-05
Maryia/build: update npm version to 2.1.0

maryia-deriv:
- bumping up the major version because the new version will contain a breaking change in the usage of exported `ChartMode` and `Views` components.


review-deriv (commented):

```
# Description
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

PR #1533, opened 2024-03-12
[DTRA] henry/dtra-781/fix: remove an extra ticks_history call when changing symbol

henry-deriv:
remove extra ticks history call when changing symbol

### DIFF
--- a/src/store/LastDigitStatsStore.ts
+++ b/src/store/LastDigitStatsStore.ts
@@ -93,9 +93,9 @@ export default class LastDigitStatsStore {
         this.updateBars();
     }
     onMasterDataUpdate({ Close, tick }: TicksStreamResponse & { Close: number }) {
-        if (!this.context || !this.mainStore.chart.currentActiveSymbol) return;
+        if (!this.context || !this.mainStore.chart.currentActiveSymbol || !this.mainStore.lastDigitStats.isVisible) return;
         this.lastTick = tick;
-        if (this.marketDisplayName !== this.lastSymbol && this.mainStore.lastDigitStats.isVisible) {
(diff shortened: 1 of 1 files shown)


review-deriv (commented):

```
# Description
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

balakrishna-deriv (approved):

src/store/LastDigitStatsStore.ts
```diff
@@ -95,7 +95,7 @@ export default class LastDigitStatsStore {
     onMasterDataUpdate({ Close, tick }: TicksStreamResponse & { Close: number }) {
         if (!this.context || !this.mainStore.chart.currentActiveSymbol) return;
         this.lastTick = tick;
-        if (this.marketDisplayName !== this.lastSymbol) {
+        if (this.marketDisplayName !== this.lastSymbol && this.mainStore.lastDigitStats.isVisible) {
```
balakrishna-deriv, line 98: Instead of this, can we add a condition at the top like the below line?

```
if (!this.mainStore.lastDigitStats.isVisible) return;
```

balakrishna-deriv: It will prevent the whole logic in `onMasterDataUpdate` if the last digit stat is not shown.

kate-deriv (approved):

balakrishna-deriv (approved):

maryia-deriv (approved):

akmal-deriv (approved):

review-deriv (commented):

```
# Description
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

PR #1534, opened 2024-03-12
[DTRA]/Ahmad/DTRA-951/Template Message For Accumulators

ahmadtaimoor-deriv:
This PR adds an message in the template modal where accumulators and digits are added . It also disable all the template other then one with ticks where applicable (accumulators and digits).

### DIFF
--- a/sass/components/_dialog.scss
+++ b/sass/components/_dialog.scss
@@ -132,6 +132,20 @@
             }
         }
     }
+    .info-footnote {
+        @include themify($themes) {
+            color: themed('ChartModeText') !important;
+        }
+        line-height: 18px;
+        font-size: 12px;
+        font-weight: 400;
+        padding: 16px;
+        text-align: center;
+
+        &--mobile {
+            font-size: 11px;
+        }
+    }
 }
 
 .tabs {
--- a/src/components/ChartMode.tsx
+++ b/src/components/ChartMode.tsx
@@ -15,7 +15,7 @@ import {
 } from './Icons';
 import Menu from './Menu';
 import Timeperiod from './Timeperiod';
-
+import InfoFootnote from './InfoFootnote';
 
 type TChartModeProps = {
     portalNodeId?: string;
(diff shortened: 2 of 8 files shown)


src/components/ChartMode.tsx
```diff
@@ -67,17 +67,18 @@ const ChartMode = ({ onChartType, onGranularity, portalNodeId }: TChartModeProps
                     </div>
                 </div>
                 {allowTickChartTypeOnly && (
-                    <div
-                        className={classNames('sc-chart-mode__section__text', {
```
henry-deriv, line 71: can you remove the styling ?

src/components/InfoFootnote.tsx
```diff
@@ -0,0 +1,25 @@
...
+type InfoFootnoteProps = {
+    className?: string;
+    isMobile?: boolean;
+    text: string;
+};
+
+const InfoFootnote = ({ className, isMobile, text }: InfoFootnoteProps) => (
+    <div
+        className={classNames('info-footnote', className, {
+            [`${className}--mobile`]: isMobile,
```
henry-deriv, line 13: when will this styling apply? im not seeing classname passed as a class anywhere

akmal-deriv (approved):

src/utils/index.ts
```diff
@@ -93,7 +93,7 @@ export const getIntervalInSeconds = ({
     timeUnit,
     interval,
 }: {
-    timeUnit?: string | number;
+    timeUnit?: string | number | null;
```
balakrishna-deriv, line 96: image

Can we add a null check here since we are adding null to the `timeUnit` type?

sass/components/_dialog.scss
```diff
@@ -132,6 +132,20 @@
             }
         }
     }
+    .info-footnote {
+        @include themify($themes) {
+            color: themed('ChartModeText') !important;
+        }
+        line-height: 18px;
+        font-size: 12px;
+        font-weight: 400;
```
kate-deriv, line 141: font-weight: 400 (aka normal) is default, it can be removed in case if it's not overwritten something. If it's for overwritten, please, ignore my comment.

src/components/InfoFootnote.tsx
ahmadtaimoor-deriv: classname is initialized with 'info-footnote' now and in future we can reuse with component with a different className (incase we need different style ) , we can just pass the classname into component that will replace 'info-footnote'.

src/utils/index.ts
ahmadtaimoor-deriv: Done now thanks

farhan-nurzi-deriv (approved):

src/components/Views.tsx
```diff
@@ -38,18 +41,22 @@ type TActiveListViewProps = {
     onGranularity: (granularity?: TGranularity) => void;
 };
 
-const ViewItem = ({ view, remove, onClick }: TViewItemProps) => (
+const ViewItem = ({ disabled, view, remove, onClick }: TViewItemProps) => (
```
henry-deriv, line 44: isDisabled would be better

maryia-deriv (dismissed):

src/utils/index.ts
```diff
@@ -110,14 +110,14 @@ export const getIntervalInSeconds = ({
         unit = 1;
     }
 
-    if (unit !== undefined && interv !== undefined) {
+    if (interv !== undefined && unit != null) {
```
balakrishna-deriv, line 113:
```suggestion
    if (unit !== undefined && interv !== undefined && unit != null) {
```

PR #1537, opened 2024-03-15
[DTRA]/Ahmad/DTRA-943/save drawing in template

ahmadtaimoor-deriv:
This PR allows to save drawing tools in the templates . So now a user can save drawing tools in addition with crosshair, indicator, chart Type and chart Interval.


akmal-deriv (approved):

kate-deriv (approved):

henry-deriv (approved):

farhan-nurzi-deriv (approved):

chart_app/lib/src/add_ons/add_ons_repository.dart
```diff
@@ -101,6 +101,52 @@ class AddOnsRepository<T extends AddOnConfig> extends ChangeNotifier
...
+    }
+    onLoadCallback?.call(_addOns.map((e) => jsonEncode(e)).toList());
+    notifyListeners();
+    _writeToPrefs(); 
+  }
+
+  ///Remove the template
+  Future<void> removeTemplate(String templateName) async {
+  if (_prefs != null && _prefs!.containsKey('template_$templateName')) {
+    await _prefs!.remove('template_$templateName');
```
maryia-deriv, line 131: just a question :) it possible to get rid of '!'s by doing the null check?

src/store/DrawToolsStore.ts
```diff
@@ -364,15 +363,33 @@ export default class DrawToolsStore {
...
+    } else {
             const item = activeTools[groupIndex];
             item.items.push({ ...finalItem, ...{ num: item.items.length } });
             activeTools[groupIndex] = item;
         }
         this.activeToolsGroup = activeTools;
     }
 
+
+    saveTemplate(name:string){
```
maryia-deriv, line 376:
```suggestion
    saveTemplate(name: string){
```
is a space supposed to be there? same for the below occurrences

maryia-deriv (changes requested):

balakrishna-deriv:
We should not be saving drawing tools in a template as the y-axis range differs for each symbol. It makes sense to save indicators in a template but not drawing tools.

PR #1545, opened 2024-03-20
Bala/update version to 2.1.7

balakrishna-deriv:
Update version to 2.1.7


review-deriv (commented):

```
# Description
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

.stylelintignore
```diff
@@ -0,0 +1 @@
+dist
```
balakrishna-deriv, line 1:
```suggestion
dist

```

review-deriv (commented):

```
# Description
Whys and Hows come here.

# Cards
* https://clickup.com/12345
* https://clickup.com/12346

```

PR #1554, opened 2024-03-25
[DTRA]/Ahmad/FEQ-1943/restrict unsupported indicators

ahmadtaimoor-deriv:
The PR restrict the unsupported indicators to which with time interval of 1 tick when the time interval is changed via selecting trade types such as accumulators, digits. It shows the same popup as the one that is shown when selected unsopported indicator and change its time interval to tick


app/index.tsx
```diff
@@ -256,9 +256,11 @@ const App = () => {
             });
         }
     };
-
+    const ref=React.useRef(null);
```
balakrishna-deriv, line 259: Doesn't look formatted. Can you check please?

henry-deriv (approved):

farhan-nurzi-deriv (approved):

henry-deriv (approved):

kate-deriv (approved):

src/store/StudyLegendStore.ts
```diff
@@ -326,14 +334,27 @@ export default class StudyLegendStore {
...
+        if (filteredItem.length > 0) {
+            filteredItem.forEach(item => {
+                this.mainStore.state.stateChange(STATE.INDICATOR_DELETED);
+                this.deleteStudyById(item.id);
+            });
+            this.mainStore.state.saveLayout();
+        }
+    }
+    savePredictionStudies() {
+        const filteredItem = this.activeItems.filter(item => item.isPrediction === true);
```
maryia-deriv, line 347:
```suggestion
        const filteredItem = this.activeItems.filter(item => item.isPrediction);
```

maryia-deriv (changes requested):

chart_app/lib/deriv_chart_wrapper.dart
```diff
@@ -366,7 +366,6 @@ class DerivChartWrapperState extends State<DerivChartWrapper> {
                           : configModel.msPerPx,
                       minIntervalWidth: _getMinIntervalWidth(),
                       maxIntervalWidth: _getMaxIntervalWidth(),
-                      dataFitPadding: _getDataFitPadding(rightPadding),
```
maryia-deriv, line 369: just a question: could you please tell more about why we need to remove it? :)
If we don't need it, it might be better to remove unused  _getDataFitPadding() declaration

ahmadtaimoor-deriv: Initially, I included the _getDataFitPadding() declaration because I encountered some errors locally that seemed to be resolved on it's removal . But it works fine on the testlink  as I checked now . I added it back .

balakrishna-deriv (dismissed):

maryia-deriv (approved):

maryia-deriv (dismissed):

balakrishna-deriv (approved):

henry-deriv (approved):

maryia-deriv (approved):

PR #1605, opened 2024-05-30
[DTRA] Maryia/DTRA-1276/fix: make chart detect every settings.language update & handle its case

maryia-deriv:
##### Before:
- the chart was detecting the `language` change in `settings` object prop only on the initial load of the chart,
- the chart was accepting the `language` value, which comes inside `settings` object prop, only as a lower-case string.
##### After:
- the chart is applying an updated `language` value, which comes inside `settings` object prop, on every update of the value,
- the chart is handling the case of the `language` value regardless of its initial case.


src/store/ChartStore.ts
```diff
@@ -272,7 +272,6 @@ class ChartStore {
         const { chartSetting } = this.mainStore;
         chartSetting.setSettings(settings);
         chartSetting.onSettingsChange = onSettingsChange;
-        localStorage.setItem('current_chart_lang', settings?.language || 'en');
```
maryia-deriv, line 275: moved it to src/store/ChartSettingStore.ts

kate-deriv (approved):

balakrishna-deriv (approved):

balakrishna-deriv (approved):

